PR #1, opened 2022-02-15
Add working draft of FiatConnect API specification

jophish:


jophish:
A brief summary of recent changes:
* Update `/transfer` endpoints response bodies to include a `transferAddress` field.
* Add data about fee types and frequency to the `/quote` endpoint response bodies
* Add data about settlement times to the `/quote` endpoint response bodies
* Add data about duration for which a quote's price is guaranteed to the `/quote` endpoint response bodies
* Add information about sandbox environments

cajubelt (approved):
at a high level the changes from you pr description look fine but I didn't look in detail. Would it make sense to commit the original first and then this one so we can see the diff more easily? Up to you, I'll approve this so you can do whatever makes sense to you

PR #5, opened 2022-03-02
Add allowedValues field for schemas in quote response

jophish:
This addition to the quote endpoint allows providers to specify a list of acceptable values that clients can submit for fields in schemas. This can be used on the client side to display a dropdown/selection for these fields, rather than have the user enter arbitrary values which might be rejected by the server, for example.

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -529,11 +529,21 @@ On success, the server MUST return an HTTP `200`, with the following response bo
 	},
 	kyc: {
 		kycRequired: `boolean`,
-		kycSchemas: `KycSchemaEnum[]`
+		kycSchemas: {
+			kycSchema: `KycSchemaEnum`,
+			allowedValues: {
+				[string]: `string[]`
+			}
+		}[]
 	},
 	fiatAccount: {
 		[FiatAccountTypeEnum]: {
-			fiatAccountSchemas: `FiatAccountSchemaEnum[]`,
+			fiatAccountSchemas: {
+				fiatAccountSchema: `FiatAccountSchemaEnum`,
+				allowedValues: {
+					[string]: `string[]`
+				}
+			}[],
 			fee?: `float`,
 			feeType?: `FeeTypeEnum`,
 			feeFrequency?: `FeeFrequencyEnum`,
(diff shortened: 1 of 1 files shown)


fiatconnect-api.md
```diff
@@ -529,11 +529,21 @@ On success, the server MUST return an HTTP `200`, with the following response bo
 	},
 	kyc: {
 		kycRequired: `boolean`,
-		kycSchemas: `KycSchemaEnum[]`
+		kycSchemas: {
+			kycSchema: `KycSchemaEnum`,
+			allowedValues: {
+				[string]: `string[]`
+			}
```
jh2oman, line 536: Two questions: 
1. Why does allowedValues live in the quote as opposed to being defined in the actual schema?
2. allowedValues seems like is solves a very specific problem with rendering a dropdown on a UI, but I can think of several others problems that we'll likely run into very soon such as different And/OR combinations, inputing dates, inputing pictures... My point is, do you think we should look into a more generic schema definition solution for handling these cases? I worry that if we try to tackle them one at a time we'll end up with a mess in the end

jophish: 1. allowedValues is dynamic; providers can choose to provide it for any schema they wish, and the contents of allowedValues is expected to change depending on various parameters (e.g., user geo). Schemas define the information that a user must send to the provider in order to register information with them; it wouldn't make sense to include allowedValues there, since it doesn't correspond to a field the user needs to provide information for. The expected flow is that the client will store the allowedValue field after calling the quote endpoint, and only allow a user to select an allowed value when providing information for the appropriate fields.
2. Yeah; you're definitely right. I think AND/OR combinations of KYC schemas is a different question entirely, but still a good point (my vote's still in favor of just ANDing schemas as a simple-ish solution). Inputting dates and pictures etc. is a much more closely related problem, though. It could make sense for FiatConnect schemas to be defined in a similar way to how `ajv` schemas are defined; e.g.:

```
{
    date: {
        type: FiatConnect.Date
    },
    name: {
        type: FiatConnect.String
    }
}
```

etc. This would be a pretty substantial change to the specification though, and require a lot of changes to the starter API repo (and likely partner implementations), so it probably warrants more discussion.

fiatconnect-api.md
```diff
@@ -672,15 +686,25 @@ On success, the server MUST return an HTTP `200`, with the following response bo
 		cryptoType: `CryptoTypeEnum`,
 		fiatAmount: `float`,
 		cryptoAmount: `float`,
-		guaranteedUntil: `string`
+		guaranteedUntil?: `string`
```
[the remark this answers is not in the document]
jophish: Yeah; I noticed that this was wrong so included it as a drive-by change..

fiatconnect-api.md
jophish: Although, thinking about the strawperson solution in my response to 2., it might actually make more sense to include `allowedValues` as an optional field within the schema itself... have to think about it some more, though.

jh2oman: 1. `allowedValues` is basically defining a string value as an enum, right? If we have types like your strawperson proposal in 2 then the fiatconnect API can just return different schemas with different enum options.
2. I think something like that will be necessary to make a UI that is both maintainable and usable. Some of the types that I think of off the top of my head that will be necessary for this are Date, Number, PhoneNumber, Enum, String, Boolean, Address, Picture, Document.

Because on the UI we probably need to:
1. Have at least a little copy to help the user (for instance we will not just display a field title like isoCountryCode to them, it will just say Country)
2. Have some basic validation (for instance insuring that their zip code is valid for their state, that their date is a valid format, that their phone number is a valid format)

In order to do those in a non-bespoke way we need much richer types than just `string` like the current [PersonalDataAndDocuments schema](https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#728-personaldataanddocuments) has for all fields

jh2oman: One alternative I see is we could be very strict on the key names for KYC / FiatAccount info. For instance enforcing that any date must have the substring `date` in it or phone numbers must always be `phoneNumber`. We could derive types from that on the client.

fiatconnect-api.md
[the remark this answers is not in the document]
cajubelt: please separate it out

fiatconnect-api.md
cajubelt: > including allowedValues in the schema itself would represent a significant shift in how we do things I think

Can you give an example of a change we would need to make @jophish ? Not sure I follow.

I like the idea of just defining schemas as you would for `ajv`

jh2oman: Original implementation sounds good 😆

jh2oman (approved):

PR #11, opened 2022-03-12
Optionally declare wallet address for transfer request payload

d-mooers:
In cases where a wallet is a smart contract, the address linked to a user's fiatConnect account will be their EOA rather than their wallet, as they will sign the jwt token with their private key for user auth. In this case when a user goes to initiate a transfer, it will transfer the funds to their EOA rather than their smart contract wallet, and if it is a transfer out the provider will be looking for transfers from the EOA rather than the smart wallet. This can be handled by the ui, but it is one additional txn for the user to sign.

The wallet field could default to the address of the fiatConnect account, so if not necessary it would not need to be provided.

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -1180,6 +1180,8 @@ The request body must contain the following fields:
   - The amount of fiat currency to transfer in.
 * `fiatAccountId`: {`string`} [REQUIRED]
   - The fiat account ID to use for the transfer.
+* `destinationAddress`: {`string`} 
+  - An optional address to receive the transfer, in place of account associated with fiatAccountId 
 
 ##### 3.3.4.1.2. Responses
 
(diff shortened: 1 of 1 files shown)


cajubelt (approved):

jh2oman (approved):

lisaqyli (approved):

cajubelt (changes requested):
concern I just thought of--

what if someone monitored the transactions to the provider's wallet address and attempted creating a transfer out for a to that address?
Something like:
- Alice initiates a transfer out with PayChant with "senderAddress" as her own wallet address
- Alice sends funds to PayChant
- Bob initiates a transfer out with PayChant with "senderAddress" as Alice's wallet address (but his own fiat account as the destination)

(using PayChant arbitrarily as an example provider here)

Seems like an attack that providers could resolve by looking at the timing of their transfer out requests, but I worry that relies too much on providers' awareness of this kind of attack.

Interested to hear what others think on this. It's possible that auth changes we are considering will render this point (and possibly the whole PR) moot by allowing smart-contract-mediated signatures, tbd

PR #16, opened 2022-03-28
Add AccountNumber Fiat Account schema for Paychant

jophish:


fiatconnect-api.md
```diff
@@ -1655,9 +1659,7 @@ represents what *kind* of account that schema represents; e.g., Debit Card, Cred
 
 ```
 [
-	`CheckingAccount`,
-	`DebitCard`,
-	`CreditCard`
+	`BankTransfer`
```
satish-ravi, line 1662: `Transfer` sounds more like an operation than an account. Does this represent any bank account that support transfers? Would checking account / debit card all fall under `BankTransfer" or are they still separate types and if they are the same, what are other account types?

Also, do the examples above need to be updated?

fiatconnect-api.md
```diff
@@ -1717,10 +1727,22 @@ A KYC schema containing personal data about a user, as well as documents such as
...
-## 7.3. Initial Entity Support
+### 7.3.2. Fiat Account Schemas
 
-This section details the initial entity support for FiatConnect. In particular, the initial KYC and Fiat Account schemas that the FiatConnect specification
-should support for initial integrators. This section is pending further research and feedback from potential CICO providers.
+All Fiat Account Schemas supported by FiatConnect MUST contain the `name`, `institution`, and `fiatAccountType` fields. `name` is a friendly, user-definable name for the account.
+`institution` is the financial institution/organization the account is with, and `fiatAccountType` is a `FiatAccountTypeEnum` value, representing the type of fiat account this
+schema represents.
+
+####  7.3.2.1. `NGBankTransfer`
```
satish-ravi, line 1736: if this is specific to Nigeria, would there be one type per country? Would that be scalable?

jh2oman: If you want to go more generic, this could just be named `AccountNumber` since all the other fields are required.

jophish: I'm somewhat concerned about the re-usability of these schemas... syntactically, this schema may be reusable for other geos. But semantically, this schema is for collecting this information for an account located in Nigeria. When a client receives this enum, it will know to prompt the user for a Nigerian account. That could mean a custom UI/integration that prompts for a Nigeria-specific account, or some restriction on what can be entered as `accountNumber` to comply with the expected syntax for Nigerian accounts.

Obviously, duplicating this schema for every country is not really scalable from a client's perspective (who would need to implement handlers for each possible schema). I'm not sure what the best approach is here. @cajubelt, any thoughts on this?

fiatconnect-api.md
jophish: I'm not entirely sure. Checking account would probably fall under `BankTransfer`, but debit card might be a different account type entirely (e.g., `BankTransfer` could represent transfer methods that use ACH, but debit card payments are available in real time so could be considered another account type). It's hard to say exactly until we get more examples of required schemas/data from providers.

fiatconnect-api.md
cajubelt: one thought that comes to mind is adding a `country` field to the schema and using `allowedValues` to restrict that to `NG` for some use cases.

That way, the client could get clued in that only Nigerian bank accounts are allowed, and maybe surface that info somehow to the user (could even have a line of text saying: "Account must be from Nigerian bank", something like that)

cajubelt: I'd also be fine with just renaming to `AccountNumber` to avoid over-engineering this.

fiatconnect-api.md
```diff
@@ -1717,10 +1727,22 @@ A KYC schema containing personal data about a user, as well as documents such as
...
+
+```
+{
+	name: `string`,
+	institution: `string`,
```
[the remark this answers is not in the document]
jophish: This is a user-friendly institution name. The idea behind this is that when returning obfuscated account information in the  GET /accounts endpoint, it should contain an institution name and "account name" within that institution that can be displayed to the user so they can distinguish between accounts.

fiatconnect-api.md
jophish: I like changing the schema name to `AccountNumber`. Having a `country` field that we restrict via `allowedValues` seems like a good option, albeit a bit hacky. I think it's the best solution we have that doesn't introduce a major redesign though, so I'm comfortable including it here.

fiatconnect-api.md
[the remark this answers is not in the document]
cajubelt: how about changing the fields to `accountName` and `institutionName`?

fiatconnect-api.md
jh2oman: How about just `BankAcount`?

jophish: `BankAccount` sounds perfect to me.

jh2oman (approved):

fiatconnect-api.md
```diff
@@ -1717,10 +1727,30 @@ A KYC schema containing personal data about a user, as well as documents such as
...
+
+`AccountNumber` is a Fiat Account schema that represents accounts where the only identifying information required is some `accountNumber` string.
+
+```
+{
+	accountName: `string`,
+	institutionName: `string`,
+	accountNumber: `string`,
+	country: `string`
+	fiatAccountType: `FiatAccountTypeEnum.BankAccount`
```
satish-ravi, line 1747: What's the relationship between fiatAccountType and fiatAccountSchema? Is this always `BankAccount` for `AccountNumber` schema or can those be different?

fiatconnect-api.md
```diff
@@ -191,6 +191,10 @@
     + [7.2.6. `FeeFrequencyEnum`](#726--feetypeenum-)
     + [7.2.7. `FiatAccountSchemaEnum`](#727--fiataccountschemaenum-)
   * [7.3. Initial Entity Support](#73-initial-entity-support)
+    + [7.3.1. KYC Schemas](#731-kyc-schemas)
+	  - [7.3.1.1. `PersonalDataAndDocuments`](#7311--personaldataanddocuments-)
+	+ [7.3.2. Fiat Account Schemas](#732-fiat-account-schemas)
+	  - [7.3.2.1. `NGBankTransfer`](#7311--ngbanktransfer-)
```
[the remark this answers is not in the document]
jophish: thanks for catching this!

fiatconnect-api.md
jophish: `AccountNumber` is a schema (potentially one of many) that can be used to transmit information about a Fiat Account of type `BankAccount`, so for `AccountNumber`, this would be fixed, in order to reify the mapping of `AccountNumber` -> `BankAccount`. This could in theory be an implicit mapping, though; we could define it somewhere in the spec by explicitly saying `AccountNumber` represents a Fiat Account of type `BankAccount`.

cajubelt: I prefer it here fwiw

fiatconnect-api.md
```diff
@@ -1717,10 +1727,30 @@ A KYC schema containing personal data about a user, as well as documents such as
...
+	country: `string`
+	fiatAccountType: `FiatAccountTypeEnum.BankAccount`
+}
+```
 
-This section details the initial entity support for FiatConnect. In particular, the initial KYC and Fiat Account schemas that the FiatConnect specification
-should support for initial integrators. This section is pending further research and feedback from potential CICO providers.
+The `country` field is meant to provide a way for providers to specify which country this data is meant for. For example, if a provider is expecting an `AccountNumber`
+schema for Nigeria, they can set the `allowedValues` field for `country` in the quote endpoint response to `['NG']`. With this information, the client will know
+to prompt the user for a Nigeria-specific account number.
```
cajubelt, line 1753:
```suggestion
The `country` field should be a [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) country code. The field is for providers to specify which country this data is meant for. For example, if a provider is expecting an `AccountNumber`
schema for Nigeria, they can set the `allowedValues` field for `country` in the quote endpoint response to `['NG']`. With this information, the client will know
to prompt the user for a Nigeria-specific account number.
```

cajubelt (approved):
need to specify the country field format. otherwise approve

PR #17, opened 2022-03-29
Rework Authentication scheme, require SIWE, remove JWT

jophish:
This PR drops support for the JWT auth scheme in favor of using the Sign-In With Ethereum standard.

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -224,6 +224,7 @@
     + [8.1.5. [EIP-55]](#815--eip-55-)
     + [8.1.6. [EIP-191]](#816--eip-191-)
     + [8.1.7. [EIP-1271]](#817--eip-1271-)
+    + [8.1.8. [EIP-2098]](#818--eip-2098-)
   * [8.2. Informative References](#82-informative-references)
     + [8.2.1. Webhook Best Practices](#821-webhook-best-practices)
     + [8.2.2. Idempotency Keys](#822-idempotency-keys)
@@ -2008,6 +2009,10 @@ EIP-191: Signed Data Standard, <https://eips.ethereum.org/EIPS/eip-191>
 ### 8.1.7. [EIP-1271]
 
 EIP-1271: Standard Signature Validation Method for Contracts, <https://eips.ethereum.org/EIPS/eip-1271>
+
+### 8.1.8. [EIP-2098]
+
+EIP-2098: Compact Signature Representation, <https://eips.ethereum.org/EIPS/eip-2098>
 
 ## 8.2. Informative References
 
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -33,6 +33,7 @@
 		    - [3.2.1.5.3.2.1. `InvalidSignature`](#3215321--invalidsignature-)
 		    - [3.2.1.5.3.2.2. `InvalidParameters`](#3215322--invalidparameters-)
 		    - [3.2.1.5.3.2.3. `ContractLoginNotSupported`](#3215323--contractloginnotsupported-)
+		    - [3.2.1.5.3.2.4. `NonceInUse`](#3215324--nonceinuse-)
 	  - [3.2.1.6. Using Sessions](#3216-using-sessions)
 	    * [3.2.1.6.1. Non-Privileged Endpoints](#32161-non-privileged-endpoints)
 	    * [3.2.1.6.2. Privileged Endpoints](#32162-privileged-endpoints)
@@ -558,21 +559,25 @@ sure that nonce management works correctly when scaling server resources.
 
 On failure, the server MUST return an HTTP `401` error, along with an error code that details exactly why the request failed.
 
-###### 3.2.1.5.3.2.1 `InvalidSignature`
+###### 3.2.1.5.3.2.1. `InvalidSignature`
 
 If the signature is invalid or doesn't match the included `address` field, the server MUST return an `InvalidSignature` error. Note that the address used
 to sign the message may not actually match the address being authenticated in the case where the client is authenticating as a contract-owned account. In
 such cases, `InvalidSignature` MUST NOT be returned when the signature is valid, and the address used to sign the message is considered valid by the contract
 referenced in the `address` field.
 
-###### 3.2.1.5.3.2.2 `InvalidParameters`
+###### 3.2.1.5.3.2.2. `InvalidParameters`
 
 If the request body is missing, malformed, or fails to pass any of the required checks, the server MUST respond with an `InvalidParameters` error.
 
-###### 3.2.1.5.3.2.3 `ContractLoginNotSupported`
+###### 3.2.1.5.3.2.3. `ContractLoginNotSupported`
 
 If the server determines that a client is trying to log in as a contrract-owned account but the server does not support contract-owned account logins, the
 server MUST respond with a `ContractLoginNotSupported` error.
+
+###### 3.2.1.5.3.2.4. `NonceInUse`
+
+If the nonce included in the message is already attached to a valid session, the server MUST respond with a `NonceInUse` error.
 
 #### 3.2.1.6. Using Sessions
 
@@ -1779,7 +1784,8 @@ An enum listing the error types used by various endpoints.
 	`SessionExpired`,
 	`SessionTooOld`,
 	`InvalidParameters`,
-	`ContractLoginNotSupported`
+	`ContractLoginNotSupported`,
+	`NonceInUse`
 ]
 ```
 
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -32,6 +32,7 @@
 		  + [3.2.1.5.3.2. Failure](#321531-failure)
 		    - [3.2.1.5.3.2.1. `InvalidSignature`](#3215321--invalidsignature-)
 		    - [3.2.1.5.3.2.2. `InvalidParameters`](#3215322--invalidparameters-)
+		    - [3.2.1.5.3.2.3. `ContractLoginNotSupported`](#3215323--contractloginnotsupported-)
 	  - [3.2.1.6. Using Sessions](#3216-using-sessions)
 	    * [3.2.1.6.1. Non-Privileged Endpoints](#32161-non-privileged-endpoints)
 	    * [3.2.1.6.2. Privileged Endpoints](#32162-privileged-endpoints)
@@ -410,7 +411,7 @@ with the provider.
 ### 3.2.1. Sign-In With Ethereum
 
 FiatConnect-compliant APIs are required to use the *Sign-In With Ethereum* standard, otherwise known as [EIP-4361](https://eips.ethereum.org/EIPS/eip-4361)
-in order to identify users and confirm ownership of a Celo blockchain address (private/public keypair). At a high level, the Sign-In With Ethereum standard
+in order to identify users and confirm ownership of a Celo blockchain address. At a high level, the Sign-In With Ethereum standard
 requires the client to use a user's private key to sign a plaintext message containing a number of standardized fields, and send that signed message to the
 server to prove account ownership. FiatConnect APIs MUST use the SIWE standard to create authenticated sessions for the user, and return session cookies to
 the client. More detail on exactly how this should be structured is provided below.
@@ -424,13 +425,14 @@ languages.
 
 At a high level, the client will craft a SIWE-compliant message and sign it with the user's private key. This message will contain various parameters relating
 to the origin URL which they're requesting to be authorized against, desired session length, etc. Once the client has crafted and signed the message, the client will send
-ithe signed message to a FiatConnect API at a `POST /auth/login` endpoint. The server will validate the signature, and check that all the included fields in the signed message are valid.
+the signed message to a FiatConnect API at a `POST /auth/login` endpoint. The server will validate the signature, and check that all the included fields in the signed message are valid.
 If everything checks out, the server will create a session for the user according to the details in the signed message. Once the server has created the session,
 it responds to the client with a `200`, and returns session cookies that the client can use in subsequent requests to privileged endpoints on behalf of the signed-in address.
 
 #### 3.2.1.3. SIWE Message Format
 
-Within the FiatConnect specification, all references to SIWE-compliant messages refer to plaintext messages adhering exactly to the following template.
+Within the FiatConnect specification, all references to SIWE-compliant messages refer to plaintext messages adhering exactly to the
+[SIWE message format](https://docs.login.xyz/general-information/siwe-overview/eip-4361#abnf-message-format), and additionally to the contraints described below.
 
 ```
 ${domain} wants you to sign in with your Ethereum account:
@@ -447,26 +449,28 @@ Expiration Time: ${expiration-time}
 Each field in this template is required to be filled in by the client, and has a unique meaning. As described later, servers MUST reject authorization attempts
 if any fields are not as expected. The meaning of each of these fields is enumerated below:
 
-The `domain` field MUST correspond to the hostname of the API that the client is requesting to be authorized for, e.g. `example.com`.
-The `address` field MUST correspond to the Celo blockchain address that the user is attempting to login/authorize themself as.
+The `domain` field MUST correspond to the [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2) authority of the API that the client is requesting to be authorized for, e.g. `example.com`.
+The `address` field MUST correspond to the Celo blockchain address that the user is attempting to login/authorize themself as. For externally-owned accounts, this
+must conform to the mixed-case checksum encoding specified in [EIP-55](https://eips.ethereum.org/EIPS/eip-55).
 The `uri` field MUST correspond to the origin URL of the API to authorize for, with the `/auth/login` path appended, e.g. `https://example.com/auth/login`.
-The `nonce` field MUST be a randomly generated, unique nonce chosen by the client, included for security purposes.
+The `nonce` field MUST be a unique 8 character alpha-numeric string not previously seen in as part of any unexpired request from the given client.
 The `issued-at` field MUST be the ISO 8601 datetime string of the time at which the client generated the message.
 The `expiration-time` field MUST be an ISO 8601 datetime string that specifies when this message will no longer valid, and when a session created with this message
 will expire. This field MUST NOT be more than four hours (14400 seconds) later than the `issued-at` field.
 
 Note that in cases where a client is attempting to authorize themselves as the owner of a contract-owned account (rather than an externally-owned account), the `address`
-field MUST correspond to the address of the on-chain contract account, rather than the externally-owned account (EOA) address. In all cases, however, the client should
-still sign the message using the EOA's private key.
+field MUST correspond to the address of the on-chain contract account, rather than the externally-owned account (EOA) address. The message itself should be signed in accordance
+with the [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) standard, in such a way that the contract account can validate the signature and message hash with its
+implementation of the `isValidSignature` method as outlined in the EIP-1271 standard. If desired, CICO providers MAY support authenticating as a contract-owned account,
+but it is not required.
 
 #### 3.2.1.4. Session Cookies
 
-Once the client generates a SIWE-compliant plaintext message, it signs it with the user's externally-owned account private key, resulting in an [EIP-191](https://eips.ethereum.org/EIPS/eip-191)
-Signed Data Standard format message. After generating the signed message, the client will send it to the server at a `POST /auth/login` endpoint, described below. If the server
-accepts the signed message as valid, it will respond with a session cookie linked that identifies the user's authenticated session within the server. The name of the session
-cookie is not mandated by this specification, but the server MUST use a consistent name for the session cookie, and SHOULD make a best-effort at choosing a unique name, that
-will not conflict with the session cookies served by other FiatConnect-compliant APIs. It is the client's responsibility to manage session cookies between different FiatConnect
-APIs.
+Once the client generates a SIWE-compliant plaintext message, it signs it with the user's externally-owned account private key according to [EIP-191](https://eips.ethereum.org/EIPS/eip-191)
+`personal_sign` format. The client will then send the message and signature to the server at a `POST /auth/login` endpoint, described below. If the server
+accepts the signed message as valid, it will respond with a session cookie linked that identifies the user's authenticated session within the server. The server
+MUST set the [SameSite](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite) policy on the session cookie to prevent the client
+from sending the cookie to other hosts. Ultimately, it is the client's responsibility to manage session cookies between different FiatConnect APIs.
 
 If a client sends a session cookie when calling the `POST /auth/login` endpoint (either to refresh their current, valid session, or to generate a new session if their current one is
 expired), the server MUST ignore the provided cookie. If the authentication request is valid, the server MUST set a new value on the cookie referencing the client's new, valid session,
@@ -475,15 +479,22 @@ and return it to the client.
 #### 3.2.1.5. `POST /auth/login`
 
 The `POST /auth/login` endpoint is responsible for verifying signed messages sent by clients, creating and authenticating a user's session, and returning session cookies.
-Note that both the Sign-In With Ethereum standard, and FiatConnect, support authorization for externally owned accounts (EOAs) and contract-owned accounts; The `POST /auth/login`
-endpoint MUST honor login requests for both, as described below.
+Note that the Sign-In With Ethereum standard, and FiatConnect, support authorization for externally owned accounts (EOAs) and contract-owned accounts; The `POST /auth/login`
+endpoint MUST honor login requests for EOAs, and MAY support smart contract-owned accounts as well, as described below.
 
 ##### 3.2.1.5.1. Parameters
 
 ###### 3.2.1.5.1.1. Request Body
 
-The *plaintext* request body for this endpoint MUST correspond to a signed SIWE message (as described above) in the [EIP-191](https://eips.ethereum.org/EIPS/eip-191)
-Signed Data Standard format, as required by the SIWE specification.
+The request body must contain the following fields:
+
+* `message`: {`string`} [REQUIRED]
+  - The plaintext SIWE message
+* `signature`: {`string`} [REQUIRED]
+  - The [EIP-191](https://eips.ethereum.org/EIPS/eip-191) signed message if logging in as EOA, or a signature according to the
+	[EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) standard, if logging in as a contract-owned account.
+
+Both fields should be encoded as base64 strings.
 
 ##### 3.2.1.5.2. Responses
 
@@ -493,7 +504,7 @@ On success, the server MUST return an HTTP `200`, update the state of the user's
 
 ###### 3.2.1.5.2.2. HTTP `401`
 
-On failure, the server MUST return an HTTP `400`, with the following response body.
+On failure, the server MUST return an HTTP `401`, with the following response body.
 
 ```
 {
@@ -511,9 +522,17 @@ and returns a `200`.
 
 On success, the server MUST respond with a `200` status code. The server MUST only respond successfully after performing the following checks:
 
-* A server MUST verify that the signature of the request body is valid and corresponds to the `address` field included in the signed message. If the signature does not
-  correspond to the address field, the server MUST assume that the value given for `address` corresponds to a contract-owned account, and attempt to verify that the
-  signature is valid according to the [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) standard. If neither of these checks pass, the server MUST reject the request.
+* A server MUST verify that the plaintext `message` adheres to the SIWE grammar/syntax (refer to the `ABNF Message Format` section within
+  [EIP-4361](https://eips.ethereum.org/EIPS/eip-4361)). If it does not, the server MUST return a 401 (`InvalidParameters`)
+
+* A server MUST determine whether the address being signed in as corresponds to an EOA or contract account. If a server determines that the address corresponds to a
+  contract-owned account and does *not* support contract-owned account login, it must return a 401 (`ContractLoginNotSupported`).
+
+  * If the address to sign in as corresponds to an EOA, the server MUST verify that the `signature` field was signed by the address attempting to be signed is and
+	is a valid signature for the `message` included in the request.
+
+  * If the address to sign in as corresponds to a contract-owned account and the server supports contract-owned account login, the server MUST verify the validity of
+	the included `signature` according to the [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) standard.
 
 * A server MUST verify that the `nonce` field does not correspond to a nonce that has already been used to authenticate an ongoing, valid session.
 
@@ -525,12 +544,13 @@ On success, the server MUST respond with a `200` status code. The server MUST on
   MUST fail the request. If the server's current timestamp is past the `expiration-time` field, the server MUST fail the request. If the `expiration-time` field is
   greater than four hours (14400 seconds) ahead of the `issued-at` field, the server MUST fail the request.
 
-If all fields are acceptable, and the message is formatted correctly, the server MUST authenticate the user's session, and return an HTTP `200`.
-The server MUST attach the fields from the signed message to the session internally, in order to access data about the user and their session in other endpoints.
+If all fields are acceptable, and the message is formatted correctly, the server SHOULD authenticate the user's session, and return an HTTP `200`.
+The server SHOULD attach the fields from the signed message to the session internally, in order to access data about the user and their session in other endpoints.
 
-Once a server authenticates a session using a particular nonce, the server MUST NOT allow any other sessions to be created using that nonce, until `expiration-time` of the session created
-using that nonce has been reached. The implementation of this is up to the server, but will require storing some global state about nonces in order to check if a client is attempting
-to sign-in using a nonce that is temporarily forbidden. Careful consideration should be taken to make sure that nonce management works correctly when scaling server resources.
+Once a server authenticates a session using a particular nonce, the server MUST NOT allow any other sessions to be created using the same nonce and address, until
+`expiration-time` of the session created using that nonce has been reached. The implementation of this is up to the server, but will require storing some global
+state about nonces in order to check if a client is attempting to sign-in using a nonce that is temporarily forbidden. Careful consideration should be taken to make
+sure that nonce management works correctly when scaling server resources.
 
 ###### 3.2.1.5.3.2. Failure
 
@@ -546,6 +566,11 @@ referenced in the `address` field.
 ###### 3.2.1.5.3.2.2 `InvalidParameters`
 
 If the request body is missing, malformed, or fails to pass any of the required checks, the server MUST respond with an `InvalidParameters` error.
+
+###### 3.2.1.5.3.2.3 `ContractLoginNotSupported`
+
+If the server determines that a client is trying to log in as a contrract-owned account but the server does not support contract-owned account logins, the
+server MUST respond with a `ContractLoginNotSupported` error.
 
 #### 3.2.1.6. Using Sessions
 
@@ -579,7 +604,7 @@ The second, and largest set of endpoints, are the ones that require a user to be
 * `POST /transfer/in`
 * `POST /transfer/out`
 
-For these endpoints, a user needs to be logged in with an authenticated session. This session may have the maximum expiration time allowed by the FiatConnect standard (four hours).
+For these endpoints, a user needs to be logged in with an authenticated session. This session MUST have an expiration time of no greater than four hours (the maximum allowed by FiatConnect).
 
 If a client attempts to access any of these endpoints without a session, the server MUST return a `401` status code with an `Unauthorized` error.
 If a client tries to acces any of these endpoints with an expired session, the server MUST respond with an HTTP `401` status code with a `SessionExpired` error.
@@ -600,6 +625,14 @@ A server MUST support API token authentication, and MAY *require* that clients i
 that clients include an API key in each request, it MUST respond to the client with an HTTP `400` error if the API key is missing from the request.
 Regardless of whether or not a server requires an API key on every request, it MUST return an HTTP `401` error if an API key is
 provided but does not correspond to any registered client; likewise it MUST return an HTTP `400` error if the API key is poorly formed.
+
+API token authentication is *completely separate* from SIWE-based auth. The priveleged endpoints listed earlier *always* require a user to have a valid session via SIWE
+authentication. API token authentication MAY be required on all endpoints, or optional. Regardless of whether or not the client provides an API token, if the client attempts
+to access a priveleged endpoint without a valid session, the server MUST reject their request.
+
+### 3.2.3. HTTPS
+
+All servers implementing a FiatConnect-compliant API MUST use HTTPS, and MUST NOT support unsecured HTTP requests.
 
 ## 3.3. Formal Specification
 
@@ -1743,7 +1776,8 @@ An enum listing the error types used by various endpoints.
 	`Unauthorized`,
 	`SessionExpired`,
 	`SessionTooOld`,
-	`InvalidParameters`
+	`InvalidParameters`,
+	`ContractLoginNotSupported`
 ]
 ```
 
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -415,10 +415,10 @@ languages.
 #### 3.2.1.2. SIWE Sessions
 
 At a high level, the client will craft a SIWE-compliant message and sign it with the user's private key. This message will contain various parameters relating
-to the origin URL which they're requesting to be authorized against, desired session length, etc. Once the client has crafted and signed the message, it will send
-it to a FiatConnect API at a `POST /auth/login` endpoint. The server will validate the signature, and check that all the included fields in the signed message are valid.
+to the origin URL which they're requesting to be authorized against, desired session length, etc. Once the client has crafted and signed the message, the client will send
+ithe signed message to a FiatConnect API at a `POST /auth/login` endpoint. The server will validate the signature, and check that all the included fields in the signed message are valid.
 If everything checks out, the server will create a session for the user according to the details in the signed message. Once the server has created the session,
-it responds to the client with a `200`, and returns session cookies that the client can use in subsequent requests to priveleged endpoints on behalf of the signed-in address.
+it responds to the client with a `200`, and returns session cookies that the client can use in subsequent requests to privileged endpoints on behalf of the signed-in address.
 
 #### 3.2.1.3. SIWE Message Format
 
@@ -448,15 +448,15 @@ The `expiration-time` field MUST be an ISO 8601 datetime string that specifies w
 will expire. This field MUST NOT be more than four hours (14400 seconds) later than the `issued-at` field.
 
 Note that in cases where a client is attempting to authorize themselves as the owner of a contract-owned account (rather than an externally-owned account), the `address`
-field MUST correspond to the address of the on-chain contract account, rather than the EOA address. In all cases, however, the client should still sign the message using
-the EOA's private key.
+field MUST correspond to the address of the on-chain contract account, rather than the externally-owned account (EOA) address. In all cases, however, the client should
+still sign the message using the EOA's private key.
 
 #### 3.2.1.4. Session Cookies
 
 Once the client generates a SIWE-compliant plaintext message, it signs it with the user's externally-owned account private key, resulting in an [EIP-191](https://eips.ethereum.org/EIPS/eip-191)
 Signed Data Standard format message. After generating the signed message, the client will send it to the server at a `POST /auth/login` endpoint, described below. If the server
 accepts the signed message as valid, it will respond with a session cookie linked that identifies the user's authenticated session within the server. The name of the session
-cookie is not specified by this specification, but the server MUST use a consistent name for the session cookie, and SHOULD make a best-effort at choosing a unique name, that
+cookie is not mandated by this specification, but the server MUST use a consistent name for the session cookie, and SHOULD make a best-effort at choosing a unique name, that
 will not conflict with the session cookies served by other FiatConnect-compliant APIs. It is the client's responsibility to manage session cookies between different FiatConnect
 APIs.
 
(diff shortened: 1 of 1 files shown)


fiatconnect-api.md
```diff
@@ -383,105 +406,241 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+requires the client to use a user's private key to sign a plaintext message containing a number of standardized fields, and send that signed message to the
+server to prove account ownership. FiatConnect APIs MUST use the SIWE standard to create authenticated sessions for the user, and return session cookies to
+the client. More detail on exactly how this should be structured is provided below.
 
-1. Associate a user's Celo blockchain address with the request, and
-2. Verify the requester's ownership of the provided address.
+#### 3.2.1.1. SIWE Sessions
 
-Servers MUST parse and recognize JWT tokens according to the following format. Servers MUST validate these tokens on every request.
+At a high level, the client will initiate a login request with a server by requesting a SIWE-compliant message to sign from a `GET /auth/login` endpoint. This
```
jophish, line 419: Rather than having the client create the message from scratch, I'm proposing a scheme where the server issues a "challenge" to the client in the form of a message to sign. This is more idiomatic, and is somewhat "baked into" the SIWE spec; the `${domain} wants you to sign in with your Ethereum account: ${address}` template is required in _all_ SIWE messages, implying that the message is meant to be issued by the `domain` itself. This also means the client and server can coordinate their clocks, which is actually pretty handy, especially considering the additional security constraints imposed on the transfer endpoints (described below) where clock coordination is relevant.

fiatconnect-api.md
```diff
@@ -383,105 +406,245 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+###### 3.2.1.3.1.1. Request Body
+
+The *plaintext* request body for this endpoint MUST correspond to a signed SIWE message (as described above) in the [EIP-191](https://eips.ethereum.org/EIPS/eip-191)
+Signed Data Standard format, as required by the SIWE specification.
+
+##### 3.2.1.3.2. Responses
+
+###### 3.2.1.3.2.1. HTTP `200`
+
+On success, the server MUST return an HTTP `200`, and update the state of the user's session to be authenticated.
```
jh2oman, line 526: The server also needs to return the session cookie, right?

fiatconnect-api.md
```diff
@@ -383,105 +406,245 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+
+If the signature is invalid or doesn't match the included `address` field, the server MUST return an `InvalidSignature` error.
+
+###### 3.2.1.3.3.2.2 `InvalidParameters`
+
+If the request body is missing, malformed, or fails to pass any of the required checks, the server MUST respond with an `InvalidParameters` error.
+
+#### 3.2.1.4. Using Sessions
+
+Once a client has established a session with a server, they can use that session (by sending the session cookie to the server) to access priveleged endpoints
```
jh2oman, line 576: What is the exact name of the cookie that the client uses to pass the session?

fiatconnect-api.md
```diff
@@ -383,105 +406,245 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+requirements. These are explained below.
+
+##### 3.2.1.4.1. Non-Privileged Endpoints
+
+The first group of endpoints are those that *do not* require the user to have a logged-in session. These are:
 
 * `GET /quote/in`
 * `GET /quote/out`
+
+Since it may be a painful user experience (or impossible) to have a user log in to a provider just to get a quote, authentication is not required for these endpoints. Of course,
```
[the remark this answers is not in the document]
cajubelt: agreed on the inconvenience though

fiatconnect-api.md
```diff
@@ -383,105 +406,245 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+clients may still access these endpoitns when logged in, but a logged-in session (or any session at all) is not required to access these endpoints.
+
+##### 3.2.1.4.2. Privileged Endpoints
+
+The second, and largest set of endpoints, are the ones that require a user to be logged in. These endpoints are the ones that deal with reading and writing user-specific data, *except*
+those that initiate transfers:
+
+* `POST /kyc/:kycSchema`
```
jh2oman, line 595: I have a question on how this auth fits into the recent Persona <> Fiatconnect planning I did. Specifically the situation where a user has not done persona yet. Here is the proposed flow:

> 1. User attempts to cash in/out and the Wallet sees that there is no KYC info for the FC parner
> 2. Wallet sends a POST to IHL saying that `[FcPartner]` needs `[KycSchema]` and IHL creates an entry in firebase for `[FcPartner]` and `[KycSchema]`.
> 3. User goes through Persona flow on their device
> 4. Webhook from Persona calls IHL, IHL updates the firebase entry for Persona with status updates, and once status === `'Approved'` calls POST /kyc/:kycSchema for all `[FcPartner]`'s in firebase that need KYC info from step 2.
> 5. Webhook from FC Patner calls IHL when the kyc status changes, IHL updates the firebase entry for `[FcPartner]`
> 6. Wallet calls IHL GET /kyc, gets the new status change, and the user can proceed with the transaction

How could IHL do step 4 with this auth?

fiatconnect-api.md
jophish: Yeah, I'll note that explicitly.

jophish: Yep, will add this.

fiatconnect-api.md
jophish: The exact name is unimportant, but a best-effort should probably be made by providers to ensure that the cookie name will not collide with those given by other providers. Will specify this in the spec.

fiatconnect-api.md
```diff
@@ -383,105 +406,245 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+Once the client receives the SIWE-compliant message, it will sign it using the user's private key, and send the signed message to the server at a `POST /auth/login` endpoint.
+The server will validate the signature, and check that all the included fields in the signed message are valid. If everything checks out, the server will create a session
+for the user according to the details in the signed message. Once the server has created the session, it responds to the client with a `200`, and returns session cookies that
+the client can use in subsequent requests to priveleged endpoints on behalf of the signed-in address.
 
-Servers MUST support the following JWT header, using an elliptic curve digital signature algorithm (ECDSA):
+#### 3.2.1.2. `GET /auth/login`
 
-```json
+FiatConnect APIs must implement a `GET /auth/login` method that clients can use to request a message to sign. This endpoint also has the important job of generating a nonce
```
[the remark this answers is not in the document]
jophish: I'm not sure I see how this relates exactly. How does setting a nonce when requesting a session impact the potential issues that idemopotency keys solve?

fiatconnect-api.md
```diff
@@ -383,105 +406,245 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+for the signed message, as a security measure to prevent replay attacks where an attacker gains access to a signed message in transit. Additionally, this endpoint allows
+client and server to coordinate their clocks to avoid confusion around session expiration times.
```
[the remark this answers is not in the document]
jophish: Yeah, agreed. I've updated this to remove any recommendation on clock syncing. If this becomes an issue, we can come up with some scheme for clock syncing outside of any of the existing endpoints.

fiatconnect-api.md
```diff
@@ -383,105 +406,245 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+
+##### 3.2.1.2.1. Parameters
+
+###### 3.2.1.2.1.1. Query Parameters
+
+* `sessionDuration`: {`string`} [REQUIRED]
+  - A string number that represents the requested login session duration, in seconds.
```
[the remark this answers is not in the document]
jophish: No longer relevant, since we've removed the double-round-trip idea.

fiatconnect-api.md
```diff
@@ -383,105 +406,245 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+address, it also performs various safety and security checks on the contents of the message. If all checks pass, the server updates the client's session to be authenticated,
+and returns a `200`.
+
+###### 3.2.1.3.3.1. Success
+
+On success, the server MUST respond with a `200` status code. The server MUST only respond successfully after performing a variety of checks.
+
+If the client did not send session cookies, or if the included cookie does not correspond to a valid session with an expected nonce already attached, the server MUST fail the
+request. A server MUST verify that the signature is valid and corresponds to the `address` field included in the signed message. A server MUST verify that the `nonce` field corresponds
+exactly to the nonce already set on the user's session. A server MUST verify that the `domain` and `uri` fields correspond exactly to the values expected for the
```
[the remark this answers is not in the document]
jophish: Yes, but like above, no longer relevant since the new plan only involves one request.

fiatconnect-api.md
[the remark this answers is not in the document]
jophish: not sure what my intention was with this; removed

fiatconnect-api.md
```diff
@@ -383,105 +406,245 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
-server MUST return an HTTP `400` error. This is required in order to prevent clients from generating excessivley long-lived JWTs that would pose a security
-risk if leaked.
+For these endpoints, a user only needs to be logged in with an authenticated session. This session may have the maximum expiration time allowed by the FiatConnect
+standard (two weeks). It does not matter how close to expiration the user's session is when accessing these endpoints. That means that a client can maintain a longer-lived session
+in order to access these endpoints.
+
+If a client attempts to access any of these endpoints without a session, the server MUST return a `401` status code with an `Unauthorized` error.
+If a client tries to acces any of these endpoints with an expired session, the server MUST respond with an HTTP `401` status code with a `SessionExpired` error.
+
+##### 3.2.1.4.3. Privileged, Time-Restricted Endpoints
```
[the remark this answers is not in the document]
jophish: Discussed offline; changing this to two categories only (authenticated and unauthenticated)

fiatconnect-api.md
```diff
@@ -383,105 +406,245 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+MUST fail the request. If the server's current timestamp is past the `expiration-time` field, the server MUST fail the request. If the `expiration-time` field is
+greater than two weeks (1209600 seconds) ahead of the `issued-at` field, the server MUST fail the request. If all fields are acceptable, and the message
+is formatted correctly, the server MUST authenticate the user's session, and return an HTTP `200`. The server MUST attach the fields from the signed message to the session internally,
+in order to access data about the user and their session in other endpoints.
+
+Once a server authenticates a session using a particular nonce, the server MUST NOT allow any other sessions to be created using that nonce, until `expiration-time` of the session created
+using that nonce has been reached. The implementation of this is up to the server, but will require storing some global state about nonces in order to check if a client is attempting
+to sign-in using a nonce that is temporarily forbidden. Careful consideration should be taken to make sure that nonce management works correctly when scaling server resources.
+
+###### 3.2.1.3.3.2. Failure
```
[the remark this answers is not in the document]
jophish: Don't think so; all the errors are basically more specific versions of the `Unauthorized` error anyways.

fiatconnect-api.md
```diff
@@ -383,105 +406,245 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
-```json
+FiatConnect APIs must implement a `GET /auth/login` method that clients can use to request a message to sign. This endpoint also has the important job of generating a nonce
+for the signed message, as a security measure to prevent replay attacks where an attacker gains access to a signed message in transit. Additionally, this endpoint allows
+client and server to coordinate their clocks to avoid confusion around session expiration times.
+
+##### 3.2.1.2.1. Parameters
+
+###### 3.2.1.2.1.1. Query Parameters
+
+* `sessionDuration`: {`string`} [REQUIRED]
```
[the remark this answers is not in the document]
jophish: No longer relevant, since this endpoint was removed.

fiatconnect-api.md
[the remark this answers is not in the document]
cajubelt: never mind 😆

fiatconnect-api.md
cajubelt: let's take this convo back to slack, getting kind of Valora-specific

fiatconnect-api.md
cajubelt: sounds like this is outdated now

cajubelt: so I'm pressing "resolve"

fiatconnect-api.md
```diff
@@ -18,34 +18,23 @@
...
-		  + [3.2.1.3.3.1. Success](#321331-success)
-		  + [3.2.1.3.3.2. Failure](#321331-failure)
-		    - [3.2.1.3.3.2.1. `InvalidSignature`](#3213321--invalidsignature-)
-		    - [3.2.1.3.3.2.2. `InvalidParameters`](#3213322--invalidparameters-)
-	  - [3.2.1.4. Using Sessions](#3214-using-sessions)
-	    * [3.2.1.4.1. Non-Privileged Endpoints](#32141-non-privileged-endpoints)
-	    * [3.2.1.4.2. Privileged Endpoints](#32142-privileged-endpoints)
-	    * [3.2.1.4.3. Privileged, Time-Restricted Endpoints](#32143-privileged--time-restricted-endpoints)
-	  - [3.2.1.5. SIWE Resources](#3215-siwe-resources)
+      - [3.2.1.2. SIWE Resources](#3212-siwe-resources)
```
cajubelt, line 21: is it just a weird looking diff or is there an indentation issue here?

fiatconnect-api.md
```diff
@@ -383,105 +399,201 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+The `address` field MUST correspond to the Celo blockchain address that the user is attempting to login/authorize themself as.
+The `uri` field MUST correspond to the origin URL of the API to authorize for, with the `/auth/login` path appended, e.g. `https://example.com/auth/login`.
+The `nonce` field MUST be a randomly generated, unique nonce chosen by the client, included for security purposes.
+The `issued-at` field MUST be the ISO 8601 datetime string of the time at which the client generated the message.
+The `expiration-time` field MUST be an ISO 8601 datetime string that specifies when this message will no longer valid, and when a session created with this message
+will expire. This field MUST NOT be more than four hours (14400 seconds) later than the `issued-at` field.
+
+Note that in cases where a client is attempting to authorize themselves as the owner of a contract-owned account (rather than an externally-owned account), the `address`
+field MUST correspond to the address of the on-chain contract account, rather than the EOA address. In all cases, however, the client should still sign the message using
```
cajubelt, line 451:
```suggestion
field MUST correspond to the address of the on-chain contract account, rather than the externally-owned account (EOA) address. In all cases, however, the client should still sign the message using
```

fiatconnect-api.md
```diff
@@ -383,105 +399,201 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+the EOA's private key.
+
+#### 3.2.1.4. Session Cookies
+
+Once the client generates a SIWE-compliant plaintext message, it signs it with the user's externally-owned account private key, resulting in an [EIP-191](https://eips.ethereum.org/EIPS/eip-191)
+Signed Data Standard format message. After generating the signed message, the client will send it to the server at a `POST /auth/login` endpoint, described below. If the server
+accepts the signed message as valid, it will respond with a session cookie linked that identifies the user's authenticated session within the server. The name of the session
+cookie is not specified by this specification, but the server MUST use a consistent name for the session cookie, and SHOULD make a best-effort at choosing a unique name, that
```
cajubelt, line 459:
```suggestion
cookie is not mandated by this specification, but the server MUST use a consistent name for the session cookie, and SHOULD make a best-effort at choosing a unique name, that
```

fiatconnect-api.md
```diff
@@ -383,105 +399,201 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+APIs.
+
+If a client sends a session cookie when calling the `POST /auth/login` endpoint (either to refresh their current, valid session, or to generate a new session if their current one is
+expired), the server MUST ignore the provided cookie. If the authentication request is valid, the server MUST set a new value on the cookie referencing the client's new, valid session,
+and return it to the client.
+
+#### 3.2.1.5. `POST /auth/login`
+
+The `POST /auth/login` endpoint is responsible for verifying signed messages sent by clients, creating and authenticating a user's session, and returning session cookies.
+Note that both the Sign-In With Ethereum standard, and FiatConnect, support authorization for externally owned accounts (EOAs) and contract-owned accounts; this endpoint
```
cajubelt, line 470:
```suggestion
Note that both the Sign-In With Ethereum standard, and FiatConnect, support authorization for externally owned accounts (EOAs) and contract-owned accounts. The `POST /auth/login` endpoint
```

fiatconnect-api.md
```diff
@@ -383,105 +399,201 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+
+This endpoint performs validation on a signed SIWE-compliant message. In addition to checking that the signature itself is valid and corresponds to the included
+address, it also performs various safety and security checks on the contents of the message. If all checks pass, the server updates the client's session to be authenticated,
+and returns a `200`.
 
-#### 3.2.1.2. JWT Payload
+###### 3.2.1.5.3.1. Success
 
-Servers MUST recognize the following *registered claims* within the JWT payload.
+On success, the server MUST respond with a `200` status code. The server MUST only respond successfully after performing the following checks.
```
cajubelt, line 504:
```suggestion
On success, the server MUST respond with a `200` status code. The server MUST only respond successfully after performing the following checks:
```

fiatconnect-api.md
```diff
@@ -383,105 +399,201 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+* A server MUST verify that the `nonce` field does not correspond to a nonce that has already been used to authenticate an ongoing, valid session.
 
-* `client`, or *client API key*
+* A server MUST verify that the `domain` and `uri` fields correspond exactly to the values expected for the server's hostname/origin.
 
-The semantics of these claims are below:
+* A server MUST verify that the `Version` and `Chain ID` lines in the signed message are exactly `1` and `42220`, respectively.
 
-##### 3.2.1.2.1. `"sub"` Claim
+* If the `issued-at` field in the signed message is before the server's current timestamp, or if the `issued-at` field is after the `expiration-time` field, the server
```
cajubelt, line 516: > If the `issued-at` field in the signed message is before the server's current timestamp

If the server and client exactly agree on the current time, then wouldn't `issued-at` always be before the server's current timestamp?

cajubelt: I'm confused by this

cajubelt: did you mean _after_ here?

cajubelt: like if the `issued-at` field is *after* the server's current timestamp, the request should fail?

cajubelt: if so, I am newly concerned about clock drift and wonder if I was wrong about generating these messages purely on the client

fiatconnect-api.md
```diff
@@ -383,105 +399,201 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
-Since the address provided in the `"sub"` claim may not correspond to the public/private keypair used to sign and verify the JWT, the
-server MUST validate that the signature does indeed correspond to the address provided in the `"sub"` claim.
+On failure, the server MUST return an HTTP `401` error, along with an error code that details exactly why the request failed.
 
-##### 3.2.1.2.3. `"exp"` Claim
+###### 3.2.1.5.3.2.1 `InvalidSignature`
 
-The `exp`, or *expiration time* claim represents the time until which the provided JWT should be accepted by the server. The `exp` claim should be provided
-as a numeric timestamp, defined as the number of seconds since Epoch. The `exp` claim MUST be present for all endpoints *except* the following:
+If the signature is invalid or doesn't match the included `address` field, the server MUST return an `InvalidSignature` error.
```
cajubelt, line 533: might be worth explaining that the included `address` field might not "match" the signature for contract-owned accounts

fiatconnect-api.md
```diff
@@ -383,105 +399,201 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+
+###### 3.2.1.5.3.2.2 `InvalidParameters`
+
+If the request body is missing, malformed, or fails to pass any of the required checks, the server MUST respond with an `InvalidParameters` error.
+
+#### 3.2.1.6. Using Sessions
+
+Once a client has established a session with a server, they can use that session (by sending the session cookie to the server) to access priveleged endpoints
+throughout the API. Certain endpoints impose different requirements on the in-use session. There are *two levels* of endpoints, with respect to their authentication
+requirements. These are explained below.
```
cajubelt, line 543:
```suggestion
requirements. These levels are explained below.
```

fiatconnect-api.md
```diff
@@ -383,105 +399,201 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
-JWT payload. For endpoints where the `exp` claim is required, the server MUST return an HTTP `400` if it is not present.
-A server MUST return an HTTP `401` error if the provided `exp` field is later than the current time with respect to Epoch. If an `exp` claim is
-included in a request, it MUST NOT exceed 24 hours from the current time with respect to Epoch. If the `exp` claim is more than 24 hours into the future, the
-server MUST return an HTTP `400` error. This is required in order to prevent clients from generating excessivley long-lived JWTs that would pose a security
-risk if leaked.
+### 3.2.2. Client API Key
 
-##### 3.2.1.2.4. `"client"` Claim
+In addition to the SIWE-based auth used to authenticate users and create sessions, the FiatConnect specification also supports a more traditional authentication standard, used
+*only* to identify specific clients with requests. Recall the webhook-based status monitoring mentioned earlier in this document. In order to support status
```
cajubelt, line 579:
```suggestion
 to identify specific clients with requests. Recall the webhook-based status monitoring mentioned earlier in this document. In order to support status
```
to say here sounds like it may be too prescriptive. If another use for it was found that didn't break the spec in some other way, I don't think we should box people in

fiatconnect-api.md
```diff
@@ -688,6 +802,8 @@ The `GET /quote/out` endpoint is used to retrieve quotes used for transfers out
   - An ISO 3166-1 country code representing the country where the quote should be requested for.
 * `region`: {`string`}
   - An optional ISO 3166-2 subdivision code representing a region within the provided country.
+* `address`: {`string`} [REQUIRED]
+  - An EIP-55 formatted address, representing the Celo address of the user to get the quote for.
```
cajubelt, line 806:
```suggestion
  - An [EIP-55](https://eips.ethereum.org/EIPS/eip-55) formatted address, representing the Celo address of the user to get the quote for.
```

fiatconnect-api.md
```diff
@@ -527,6 +639,8 @@ used for the transfer.
   - An ISO 3166-1 country code representing the country where the quote should be requested for.
 * `region`: {`string`}
   - An optional ISO 3166-2 subdivision code representing a region within the provided country.
+* `address`: {`string`} [REQUIRED]
+  - An [EIP-55](https://eips.ethereum.org/EIPS/eip-55) formatted address, representing the Celo address of the user to get the quote for.
```
cajubelt, line 643: for contract-owned accounts should this be the EOA or their contract address? the contract, right?

cajubelt (commented):
really good! I left a bunch of tiny things, plus one meatier question about clock drift in obtaining the session cookie

fiatconnect-api.md
jophish: Yes, I believe I meant _after_ here. Clock synchronization does seem extremely important here, though I think it's out of scope for the auth update. We should probably include some separate method by which clients and servers sync on clock in a separate PR. Probably some endpoint that returns the server's clock... although, the server's clock is typically returned in the `date` header, anyways. Regardless, the exact solution for clock synchronization should be a separate ticket IMO.

fiatconnect-api.md
jophish: Sure, I think that makes sense!

fiatconnect-api.md
jophish: Yep, clarified!

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+The `POST /auth/login` endpoint is responsible for verifying signed messages sent by clients, creating and authenticating a user's session, and returning session cookies.
+Note that both the Sign-In With Ethereum standard, and FiatConnect, support authorization for externally owned accounts (EOAs) and contract-owned accounts; The `POST /auth/login`
+endpoint MUST honor login requests for both, as described below.
+
+##### 3.2.1.5.1. Parameters
+
+###### 3.2.1.5.1.1. Request Body
+
+The *plaintext* request body for this endpoint MUST correspond to a signed SIWE message (as described above) in the [EIP-191](https://eips.ethereum.org/EIPS/eip-191)
+Signed Data Standard format, as required by the SIWE specification.
```
nategraf, line 486: So neither SIWE nor EIP-191 actually specify how to package the signed message (i.e. I'm really not sure what you are referring to as the "Signed Data Standard format"). We need to specify that here. Two options are

1. Specify that the body is the signed message data and include the signature in an `Authorization` header.
2. Package both into the body as a JSON struct with `{ message, signature }`.

In both cases, we need to specify the encodings (e.g. `base64`) and whatnot.

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
 
-Servers MUST recognize the following *registered claims* within the JWT payload.
+On success, the server MUST respond with a `200` status code. The server MUST only respond successfully after performing the following checks:
 
-* `exp`, or *expiration time*
-* `sub`, or *subject*
-* `iss`, or *issuer*
+* A server MUST verify that the signature of the request body is valid and corresponds to the `address` field included in the signed message. If the signature does not
+  correspond to the address field, the server MUST assume that the value given for `address` corresponds to a contract-owned account, and attempt to verify that the
+  signature is valid according to the [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) standard. If neither of these checks pass, the server MUST reject the request.
```
nategraf, line 516: Specifying that EIP-1271 seems to contradict what I read earlier.

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+###### 3.2.1.5.3.1. Success
 
-Servers MUST recognize the following *registered claims* within the JWT payload.
+On success, the server MUST respond with a `200` status code. The server MUST only respond successfully after performing the following checks:
 
-* `exp`, or *expiration time*
-* `sub`, or *subject*
-* `iss`, or *issuer*
+* A server MUST verify that the signature of the request body is valid and corresponds to the `address` field included in the signed message. If the signature does not
+  correspond to the address field, the server MUST assume that the value given for `address` corresponds to a contract-owned account, and attempt to verify that the
```
nategraf, line 515: I would argue against including this method of determining whether a given address is an EOA or a smart contract as a MUST. There are multiple ways to figure out whether a given address is a smart contract, and we can recommend a method, but I don't think we need to require this as the method of choice here. This method is somewhat suboptimal compare to having some kind of an indexer that can tell you before you attempt to verify the signature whether the given address is a smart contract.

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
-#### 3.2.1.2. JWT Payload
+###### 3.2.1.5.3.1. Success
 
-Servers MUST recognize the following *registered claims* within the JWT payload.
+On success, the server MUST respond with a `200` status code. The server MUST only respond successfully after performing the following checks:
 
-* `exp`, or *expiration time*
-* `sub`, or *subject*
-* `iss`, or *issuer*
+* A server MUST verify that the signature of the request body is valid and corresponds to the `address` field included in the signed message. If the signature does not
```
nategraf, line 514: We should make it more clear that the they should start by making sure the whole message is valid according to the ABNF specified in SIWE.

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
-##### 3.2.1.2.1. `"sub"` Claim
+* If the `issued-at` field in the signed message is after the server's current timestamp, or if the `issued-at` field is after the `expiration-time` field, the server
+  MUST fail the request. If the server's current timestamp is past the `expiration-time` field, the server MUST fail the request. If the `expiration-time` field is
+  greater than four hours (14400 seconds) ahead of the `issued-at` field, the server MUST fail the request.
 
-The `sub`, or *subject* claim represents the user's Celo network address. This is a required claim. If it is missing, the server MUST respond
-to the client with an HTTP `400` error. This claim SHOULD be interpreted and used within server endpoints as a unique user identifier, and as the
-source of/destination for crypto funds during transfers. The address should be formatted as 42 hexadecimal characters prefixed by the string literal
-`0x`. If the `sub` claim does not match this format, the server MUST respond to the client with an HTTP `400` error.
+If all fields are acceptable, and the message is formatted correctly, the server MUST authenticate the user's session, and return an HTTP `200`.
```
nategraf, line 528:
```suggestion
If all fields are acceptable, and the message is formatted correctly, the server SHOULD authenticate the user's session, and return an HTTP `200`.
```
Realistically, servers may still chose to reject the session (e.g. due to restrictions on where the request is coming from) and this is not something we can stop them from doing.

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+The server MUST attach the fields from the signed message to the session internally, in order to access data about the user and their session in other endpoints.
```
nategraf, line 529:
```suggestion
The server SHOULD attach the fields from the signed message to the session internally, in order to access data about the user and their session in other endpoints.
```
Since, if the server found another way to accomplish state persistance, this would not effect the client, I would not make this a MUST. In particular, it's likely providers may package session details into a JWT and put it in the client's cookies, which achieves the same goals.

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
 
-##### 3.2.1.2.2. `"iss"` Claim
+Once a server authenticates a session using a particular nonce, the server MUST NOT allow any other sessions to be created using that nonce, until `expiration-time` of the session created
```
nategraf, line 531:
```suggestion
Once a server authenticates a session using a particular nonce, the server MUST NOT allow any other sessions to be created using the same nonce and address, until `expiration-time` of the session created
```
Servers may segment nonces by client without loosing security. Obviously blocking repeated nonce globally trivially achieves this as well, so maybe not worth pointing out. The one reason this could matter is if clients use timestamp based nonces.

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
 
-The `iss`, or *subject* claim represents the user's Celo address public key, dervied from their private key. This is a required claim.
-The public key MUST be in compressed form (see [here](https://docs.ethers.io/v5/api/utils/signing-key/) for more details), and exactly 33 bytes
-long.
+###### 3.2.1.5.3.2. Failure
 
-If it is missing, the server MUST respond to the client with an HTTP `400` error. The server MUST use this claim to verify the JWT signature.
-Since the address provided in the `"sub"` claim may not correspond to the public/private keypair used to sign and verify the JWT, the
-server MUST validate that the signature does indeed correspond to the address provided in the `"sub"` claim.
+On failure, the server MUST return an HTTP `401` error, along with an error code that details exactly why the request failed.
```
nategraf, line 537: Returning `401` instead of `400` seems like a good call, but does contradict what was said above. Also, saying that the MUST return a 4XX error code on failure would be broken by server errors which should return a 5XX error.

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
 
-##### 3.2.1.2.3. `"exp"` Claim
+###### 3.2.1.5.3.2.1 `InvalidSignature`
 
-The `exp`, or *expiration time* claim represents the time until which the provided JWT should be accepted by the server. The `exp` claim should be provided
-as a numeric timestamp, defined as the number of seconds since Epoch. The `exp` claim MUST be present for all endpoints *except* the following:
+If the signature is invalid or doesn't match the included `address` field, the server MUST return an `InvalidSignature` error. Note that the address used
+to sign the message may not actually match the address being authenticated in the case where the client is authenticating as a contract-owned account. In
+such cases, `InvalidSignature` MUST NOT be returned when the signature is valid, and the address used to sign the message is considered valid by the contract
+referenced in the `address` field.
```
nategraf, line 544: I found the wording of this paragraph a bit confusing. It seems to say, "You must do X. Expect actually Y may be true. So you should actually do Z." Maybe it could be be refactored to be a bit more straightforward.

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
 * `GET /kyc/:kycSchema/status`
+* `DELETE /kyc/:kycSchema`
+* `POST /accounts/:fiatAccountSchema`
+* `GET /accounts`
+* `DELETE /account/:fiatAccountId`
 * `GET /transfer/:transferId/status`
+* `POST /transfer/in`
+* `POST /transfer/out`
+
+For these endpoints, a user needs to be logged in with an authenticated session. This session may have the maximum expiration time allowed by the FiatConnect standard (four hours).
```
nategraf, line 582:
```suggestion
For these endpoints, a user needs to be logged in with an authenticated session.
This session MUST have an expiration time of no greater than four hours.
```

Just to make it a bit more clear that this is a requirement (I also foresee service not honoring it, but what can you do 🤷‍♂️)

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+to know where to send webhook-based status updates, a server may allow clients to register an API key. The exact mechanism by which servers allocate API keys
+to clients and allow them to register webhook URLs is out of scope of this document.
 
-In addition to public-private key authentication, servers must also support API token based authentication. Recall the webhook-based status monitoring
-mentioned earlier in this document. In order to support status monitoring via webhooks, individual clients will need to be able to register a URL pointing
-to an API able to handle webhook updates from the server. Once a client has registered a webhook URL with the provider, the client needs a way to identify
-itself to the server. To uniquely identify clients in order to know where to send webhook-based status updates, a server may allow clients to register an API key.
-The exact mechanism by which servers allocate API keys to clients and allow them to register webhook URLs is out of scope of this document.
+Once a client has registered an API key with a provider and associated it with a webhook URL, a client MAY provide it to a server through an `Authorization` header,
```
nategraf, line 596: It's unclear to me how the session cookie and API key interact? If a user provides an API key, but no session cookie, does that authenticate them?

nategraf (commented):
* We should add somewhere that providers MUST use HTTPS. In our implementation, and in any client spec if it exists, we should ensure the client aborts if the provider is not using HTTPS.
* Contract signatures are underspecified, and seem to be contradictory in a couple of places. That needs to be fleshed out more.

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+If everything checks out, the server will create a session for the user according to the details in the signed message. Once the server has created the session,
+it responds to the client with a `200`, and returns session cookies that the client can use in subsequent requests to privileged endpoints on behalf of the signed-in address.
 
-```json
+#### 3.2.1.3. SIWE Message Format
+
+Within the FiatConnect specification, all references to SIWE-compliant messages refer to plaintext messages adhering exactly to the following template.
+
+```
+${domain} wants you to sign in with your Ethereum account:
```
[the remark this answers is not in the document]
cajubelt: +1 on dealing with this in the presentation layer and not changing the message that gets signed (to maintain compatibility)

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
 with the provider.
 
-### 3.2.1. JWT Authentication
+### 3.2.1. Sign-In With Ethereum
 
-Only users with a registered address on the Celo blockchain may make requests to a FiatConnect API. The FiatConnect API specification
-requires servers to recognize a JWT token signed with the user's private key, and verifiable with the user's public key. With this authentication
-scheme, a server can:
+FiatConnect-compliant APIs are required to use the *Sign-In With Ethereum* standard, otherwise known as [EIP-4361](https://eips.ethereum.org/EIPS/eip-4361)
+in order to identify users and confirm ownership of a Celo blockchain address (private/public keypair). At a high level, the Sign-In With Ethereum standard
```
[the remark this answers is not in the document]
jophish: Yep, removed this.

fiatconnect-api.md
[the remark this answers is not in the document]
jophish: ++1 on leaving this as-is and presenting it to the user in an understandable way

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+Chain ID: 42220
+Nonce: ${nonce}
+Issued At: ${issued-at}
+Expiration Time: ${expiration-time}
+```
+
+Each field in this template is required to be filled in by the client, and has a unique meaning. As described later, servers MUST reject authorization attempts
+if any fields are not as expected. The meaning of each of these fields is enumerated below:
+
+The `domain` field MUST correspond to the hostname of the API that the client is requesting to be authorized for, e.g. `example.com`.
```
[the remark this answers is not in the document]
jophish: Great callout; thanks for the specificity on this.

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+The `address` field MUST correspond to the Celo blockchain address that the user is attempting to login/authorize themself as.
```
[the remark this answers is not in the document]
jophish: Great callout!

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+The `uri` field MUST correspond to the origin URL of the API to authorize for, with the `/auth/login` path appended, e.g. `https://example.com/auth/login`.
+The `nonce` field MUST be a randomly generated, unique nonce chosen by the client, included for security purposes.
+The `issued-at` field MUST be the ISO 8601 datetime string of the time at which the client generated the message.
+The `expiration-time` field MUST be an ISO 8601 datetime string that specifies when this message will no longer valid, and when a session created with this message
+will expire. This field MUST NOT be more than four hours (14400 seconds) later than the `issued-at` field.
+
+Note that in cases where a client is attempting to authorize themselves as the owner of a contract-owned account (rather than an externally-owned account), the `address`
+field MUST correspond to the address of the on-chain contract account, rather than the externally-owned account (EOA) address. In all cases, however, the client should
+still sign the message using the EOA's private key.
```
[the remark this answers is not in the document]
jophish: I see; so the discrepancy here is that EIP-1271 does not mandate a particular signing method; just that the contract account is able to validate the signature with its `isValidSignature` method. I've updated this to be less prescriptive about the signing method for contract accounts, and that it's an optional feature rather than mandatory.

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+
+#### 3.2.1.4. Session Cookies
+
+Once the client generates a SIWE-compliant plaintext message, it signs it with the user's externally-owned account private key, resulting in an [EIP-191](https://eips.ethereum.org/EIPS/eip-191)
+Signed Data Standard format message. After generating the signed message, the client will send it to the server at a `POST /auth/login` endpoint, described below. If the server
```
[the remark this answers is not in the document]
jophish: Thanks for the clarification here!

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+accepts the signed message as valid, it will respond with a session cookie linked that identifies the user's authenticated session within the server. The name of the session
+cookie is not mandated by this specification, but the server MUST use a consistent name for the session cookie, and SHOULD make a best-effort at choosing a unique name, that
+will not conflict with the session cookies served by other FiatConnect-compliant APIs. It is the client's responsibility to manage session cookies between different FiatConnect
+APIs.
```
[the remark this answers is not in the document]
jophish: great callout, noted!

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+
+If a client sends a session cookie when calling the `POST /auth/login` endpoint (either to refresh their current, valid session, or to generate a new session if their current one is
+expired), the server MUST ignore the provided cookie. If the authentication request is valid, the server MUST set a new value on the cookie referencing the client's new, valid session,
+and return it to the client.
+
+#### 3.2.1.5. `POST /auth/login`
+
+The `POST /auth/login` endpoint is responsible for verifying signed messages sent by clients, creating and authenticating a user's session, and returning session cookies.
+Note that both the Sign-In With Ethereum standard, and FiatConnect, support authorization for externally owned accounts (EOAs) and contract-owned accounts; The `POST /auth/login`
+endpoint MUST honor login requests for both, as described below.
```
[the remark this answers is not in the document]
jophish: (smart contract login support has been changed to `MAY`)

fiatconnect-api.md
```diff
@@ -391,105 +407,204 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+
+##### 3.2.1.5.1. Parameters
+
+###### 3.2.1.5.1.1. Request Body
+
+The *plaintext* request body for this endpoint MUST correspond to a signed SIWE message (as described above) in the [EIP-191](https://eips.ethereum.org/EIPS/eip-191)
```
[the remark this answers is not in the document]
jophish: Hm, I think this should be a JSON with `message` and `signature` fields, as you mention below.

fiatconnect-api.md
jophish: Yeah, changed the previous text to align with using the EIP-1271 standard.

fiatconnect-api.md
jophish: Yeah, you're definitely right; this isn't something that we should be prescriptive about. Will just say that servers MUST figure out what kind of account the client is attempting to sign is as; *how* they do that is up to them.

fiatconnect-api.md
jophish: Good callout; specified this!

fiatconnect-api.md
jophish: Good catch; changed to `SHOULD`.

fiatconnect-api.md
jophish: Hm, not sure where I specify anything but a `401` in this section. And yeah, a `4xx` would be broken by `5xx`s, but that's true of every endpoint; not sure it's worth making an allowance for that explicitly in the spec.

fiatconnect-api.md
jophish: Yeah, refactored this section to be more clear about EOA vs. contract-owned account logins.

fiatconnect-api.md
jophish: No, they're totally separate. The API token is merely used to identify a request with a particular client (specifically, for the purpose of knowing where to send webhook events). I've clarified this in the doc a bit.

fiatconnect-api.md
```diff
@@ -391,105 +408,238 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+
+###### 3.2.1.5.2.2. HTTP `401`
+
+On failure, the server MUST return an HTTP `401`, with the following response body.
+
+```
 {
-	"alg": "ES256",
-	"typ": "JWT"
+	error: `ErrorEnum`,
```
satish-ravi, line 513: Would this just include a single enum, or do we also want to have a more descriptive error? E.g., which parameter is missing in case of `InvalidParameters`

jh2oman: oooo the UI would love to know which parameter is at fault ❤️

jophish: None of these SIWE parameters are actually chosen by the user; this is all done under the hood -- what would we display to the user if e.g., the nonce was wrong? Although for the client's sake, it might make sense to have more descriptive error types here.

fiatconnect-api.md
```diff
@@ -1614,8 +1774,12 @@ An enum listing the error types used by various endpoints.
 	`ResourceExists`,
 	`ResourceNotFound`,
 	`TransferNotAllowed`,
-	`KycExpired`
-	`InvalidParameters`
+	`KycExpired`,
+	`Unauthorized`,
+	`SessionExpired`,
+	`SessionTooOld`,
```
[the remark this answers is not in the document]
jophish: nothing... good catch -- `SessionTooOld` shouldn't be here!

fiatconnect-api.md
jophish: Just added a couple of extra error types... if there's more that would be useful to have, let me know and I can add them in!

fiatconnect-api.md
```diff
@@ -391,105 +411,251 @@ The FiatConnect API specifies *two* types of authentication; the first authentic
...
+referenced in the `address` field.
+
+###### 3.2.1.5.3.2.2. `InvalidParameters`
+
+If the request body is missing, malformed, or fails to pass any of the required checks, the server MUST respond with an `InvalidParameters` error. If another
+error type is more descriptive of the actual issue, that error should be used instead.
+
+###### 3.2.1.5.3.2.3. `ContractLoginNotSupported`
+
+If the server determines that a client is trying to log in as a contrract-owned account but the server does not support contract-owned account logins, the
```
cajubelt, line 578:
```suggestion
If the server determines that a client is trying to log in as a contract-owned account but the server does not support contract-owned account logins, the
```

fiatconnect-api.md
```diff
@@ -1602,6 +1776,10 @@ An enum listing the error types used by various endpoints.
 
 ```
 [
+	`InvalidAddress`,
```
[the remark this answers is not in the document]
jophish: Yeah; `InvalidAddress` doesn't need to exist -- it's actually not referenced anywhere.

fiatconnect-api.md
```diff
@@ -1602,6 +1776,10 @@ An enum listing the error types used by various endpoints.
...
+	`InvalidDuration`,
+	`DurationTooLong`,
```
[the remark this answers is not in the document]
jophish: Ah, this is the same as above -- `DurationTooLong` isn't referenced anywhere; `ExpirationTooLong` should be used instead.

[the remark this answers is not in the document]
jophish: `InvalidDuration` also isn't referenced anywhere... will remove that as well.

PR #19, opened 2022-04-01
feat: quoteId

cajubelt:
At Kaala Wallet's request, we added a `quoteId` field to the quote endpoint response types, and for the transfer endpoint request types. 

This means a wallet should be able to show an end-user a fiat-to-crypto exchange rate that the provider has committed to when they initiate a transfer. Without it, rates and fees would need to be estimated until the transfer clears.

For a provider, this means that an identifier for a quote must be issued alongside the quote. When a wallet tries to create a request with this `quoteId`, if the quote has not yet expired, the provider must honor the exchange rate and fees from the original quote.

NOTE: `guaranteedUntil` is now a required field for Quote responses (otherwise offering quoteId's would be of little use)


fiatconnect-api.md
```diff
@@ -597,9 +600,8 @@ receive by providing `fiatAmount` worth of the fiat currency. If `cryptoAmount`
 of fiat currency required in order to receive the requested amount of crypto.
 
 The `quote.fiatType`, `quote.cryptoType`, `quote.fiatAmount`, and `quote.cryptoAmount` fields in the response body MUST correspond to the query parameters provided to the endpoint.
-The `quote.guaranteedUntil` field represents the time that the quote is guaranteed until, as a UNIX timestamp. A server MAY choose to provide this value. If it does not, the client should assume that
-the server makes no guarantees about the actual conversion rate when performing a transfer for the given quote. If a server provides this value in the response, it MUST honor the
-provided conversion rate when initiating transfers until the time indiciated in the `quote.guaranteedUntil` field.
+The `quote.guaranteedUntil` field represents the time that the quote is guaranteed until, as a UNIX timestamp. A server MUST provide this value, and MUST honor the
```
jophish, line 603: Can you add something in here explaining the `quoteId` field in the response? e.g. "The `quoteId` field in the response body is a globally unique identifier for the returned quote, that the client uses to initiate a transfer with a guaranteed currency conversion rate" or something to that effect.

jophish: Also, it could be a good idea to explicitly require servers to maintain a history of quotes, even after their `guaranteedUntil` time. Without this, a server could throw out a `quoteId` that was past its `guaranteedUntil` time, rendering the `QuoteExpired` response code from the transfer endpoint useless.

fiatconnect-api.md
```diff
@@ -762,7 +764,7 @@ of fiat currency the user should expect to recieve in exchange for `cryptoAmount
 The `quote.fiatType`, `quote.cryptoType`, `quote.fiatAmount`, and `quote.cryptoAmount` fields in the response body MUST correspond to the query parameters provided to the endpoint.
 The `quote.guaranteedUntil` field represents the time that the quote is guaranteed until, as a UNIX timestamp. A server MAY choose to provide this value. If it does not, the client should assume that
 the server makes no guarantees about the actual conversion rate when performing a transfer for the given quote. If a server provides this value in the response, it MUST honor the
-provided conversion rate when initiating transfers until the time indiciated in the `quote.guaranteedUntil` field.
+provided conversion rate when initiating transfers until the time indicated in the `quote.guaranteedUntil` field.
```
jophish, line 767: (I know this was just a typo fix, but leaving this here anyways.) Even though we don't actually have any way to verify this, it may also make sense to require servers to return a `quoteId` for transfers out. While we have no way of knowing if servers are actually honoring conversion rates for these transfers, it may result in some providers actually enforcing guarantees for transfers out, and it makes the responses for quotes in and out uniform, which is a plus.

jophish: Actually, looking further down, it looks like you updated the spec for both `/transfer/in` and `/transfer/out`, so somewhat confused what the intended behavior is.

fiatconnect-api.md
```diff
@@ -1252,7 +1256,8 @@ return an HTTP `400` error code, along with the following response body:
 
 ###### 3.3.4.1.2.3. HTTP `404`
 
-If the selected fiat account does not exist, the server MUST respond with an HTTP `404` error code, along with the following response body:
+If the selected fiat account does not exist, or if a quote with the quote ID given in the request does not exist,
```
jophish, line 1259: The equivalent `HTTP 404` section under `/transfer/out` hasn't been updated to match this; if the intention is to require quote IDs for both transfers in and transfer out, you should update this section there as well.

fiatconnect-api.md
```diff
@@ -1287,13 +1294,18 @@ This endpoint may fail for a number of reasons; the error code returned by the e
...
+
+###### 3.3.4.1.3.2.3. `TransferNotAllowed`
 
 If a transfer is not allowed for a generic reason (such as unacceptable transfer parameters) the server MUST reject the transfer and return a `TransferNotAllowed` error.
 
-###### 3.3.4.1.3.2.3. `ResourceNotFound`
+###### 3.3.4.1.3.2.4. `ResourceNotFound`
 
-If the selected `fiatAccountId` is not found for the current user, the server MUST reject the transfer and return a `ResourceNotFound` error.
+If the selected `fiatAccountId` is not found for the current user, or if a quote with `quoteId` does not exist,
```
jophish, line 1307: Same comment as above here.

fiatconnect-api.md
cajubelt: I'd like to avoid requiring providers to maintain a history of expired quotes, so I updated the errors to account for this a bit better

cajubelt: It's a good point that a QuoteExpired error doesn't mean much if there's no instructions on how long to keep the quotes for

fiatconnect-api.md
cajubelt: I mistakenly left out transfers out, meant to include both

fiatconnect-api.md
cajubelt: I updated to use `InvalidQuote` errors for both missing and expired quotes, so providers could delete expired quotes whenever they want

jophish (approved):
One small comment, but looks great!

fiatconnect-api.md
```diff
@@ -1269,7 +1277,9 @@ respect to idempotency key errors.
 
 This endpoint allows a user to initiate a new transfer in request. The server MUST support idempotency keys, and MUST NOT accept any requests which lack them.
 If a user provides a `fiatAccountId` that refers to an account they have on file that is allowed for the transfer, and the transfer parameters are acceptable,
-and the user has non-expired KYC on file, the server MUST respond with an HTTP `200` and initiate the transfer. When a new transfer is initiated, the server MUST
+and the user has non-expired KYC on file, and a quote with a matching `quoteId` exists and has not expired, the server MUST respond with an HTTP `200` and initiate the transfer.
```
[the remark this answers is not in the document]
cajubelt: I added "for the user" language to try to account for this-- though it seems low-risk

andreujuanc:
I'm a month late, but just for the record, GET requests should not modify state. There will be problems with cache, and other server issues.

PR #22, opened 2022-04-12
added schema for inr and eur for fiat-connect alpha

SonalBhanawat:


jophish (commented):
Thanks for the PR! Left a couple of comments. In general, Fiat Account schemas should only contain the minimal amount of information required to initiate a transfer to/from an end user's fiat account. It looks like these schemas contain a lot of information that relates to KYC/user PII that would be a better fit as part of a KYC Schema. Additionally, it looks like the schemas contain some transfer-specific information, such as `sourceCurrency` and `destinationCurrency`, which will vary from transfer-to-transfer and shouldn't live in a Fiat Account schema.

fiatconnect-api.md
```diff
@@ -1794,6 +1794,71 @@ is below:
 
 * `'NG'`: The account number should be exactly 10 digits long, and only include the numbers 0-9.
 
+####  7.3.2.2. `CheckingAccountInr`
+
+`CheckingAccountInr` is a Fiat Account schema that represents accounts for INR transactions.
```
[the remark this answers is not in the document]
SonalBhanawat: https://github.com/fiatconnect/specification/pull/26

fiatconnect-api.md
```diff
@@ -1794,6 +1794,71 @@ is below:
...
+```
+{
+    depositNumber: `string`, 
+    destinationCurrency: `string`, 
+    bank_name:`string`, 
+    iban: `string`, 
+    name: `string`, 
+    address: `string`, 
+    city: `string`,
+    country: `string`, ,
```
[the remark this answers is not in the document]
SonalBhanawat: fixed

fiatconnect-api.md
```diff
@@ -1794,6 +1794,71 @@ is below:
 
 * `'NG'`: The account number should be exactly 10 digits long, and only include the numbers 0-9.
 
+####  7.3.2.2. `CheckingAccountInr`
+
+`CheckingAccountInr` is a Fiat Account schema that represents accounts for INR transactions.
+
+```
```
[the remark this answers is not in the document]
msshahanshah: The situation that you are mentioning is try when the person is withdrawing funds to his own account i.e. the sender and the beneficiary are the same. 
We will ned this info if the Sender and beneficiary are not the same.

fiatconnect-api.md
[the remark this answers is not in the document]
SonalBhanawat: Also creating another PR for `NGN` support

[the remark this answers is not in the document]
SonalBhanawat: #26 - also added NGN to this list

fiatconnect-api.md
```diff
@@ -1794,6 +1794,71 @@ is below:
...
+    dateOfBirth: `string`, 
+    email: `string`, 
+    firstName: `string`, ,
+    gender: `string`, ,
+    lastName: `string`, ,
+    mobilePhoneCountry: `string`, ,
+    mobilePhoneNumber: `string`, ,
+    senderType: `string`, , 
+    zipcode: `string`, , 
+    sourceCurrency: `string`,
```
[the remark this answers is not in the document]
SonalBhanawat: removed

fiatconnect-api.md
```diff
@@ -1794,6 +1794,71 @@ is below:
 
 * `'NG'`: The account number should be exactly 10 digits long, and only include the numbers 0-9.
 
+####  7.3.2.2. `CheckingAccountInr`
```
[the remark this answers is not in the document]
SonalBhanawat: added

fiatconnect-api.md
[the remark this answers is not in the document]
jophish: My main concern here is that this sort KYC-related information doesn't really relate to bank account information; FiatConnect doesn't natively support sending money to or from an account that you don't own. In order to do that, you would presumably have to undergo KYC for the _other_ person's account, which isn't possible. Including all of these fields also makes this schema difficult to re-use. Could we instead use a schema like this, and only allow people to cash out to accounts that they own?

```
{
    "accountName": `string`,
    "institutionName": `string`,
    "institutionCode": `string`,
    "accountNumber": `string`,
    "fiatAccountType": `FiatAccountTypeEnum.BankAccount` ,
    "providerCode": `string`
}
```
Also, how is the `providerCode` field different from the `institutionCode` field? What does the `senderType` field represent?

Also, if a user submits a valid fiat account for, say, country `A` using this schema, that account _won't_ be able to be used for a transfer in, say, `country B`, correct? In that case, you should have a separate schema for each country. In particular, if the `quote` endpoint allows a user to submit a particular Fiat Account schema, say, `FiatAccountSchemaA` to be used for a transfer, the fiat account that that user submits should be able to be re-used for _all_ other quote requests that accept `FiatAccountSchemaA`. 

With the current state of this schema, a user could cash out in country `A`, and provide this `CheckingAccount` schema for a fiat account and successfully cash out. When that user goes to country `B`, the quote endpoint will say that they can use a `CheckingAccount` schema to cash out. The client will list all their available `CheckingAccount` fiat accounts that they've already added with the provider, and the user selects the `CheckingAccount` they used from the previous transfer. When they go to cash out, the transfer will fail, since that `CheckingAccount` isn't _actually_ usable for this transfer in a different country. To avoid this, you should have a separate schema for each country.

fiatconnect-api.md
```diff
@@ -1796,6 +1796,293 @@ is below:
...
+    "institutionName": `string`,
+    "institutionCode": `string`,
+    "accountNumber": `string`,
+    "fiatAccountType": `FiatAccountTypeEnum.BankAccount`
+}
+```
+
+The `country` field should be a [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) country code. The field is for providers to specify which country this data is meant for.
+
+Based on the country `institutionName` and `institutionCode` should be supplied. AllowedValues Enum is mentioned in section 7.3.2.2.1
```
jophish, line 1815: As we discussed, could we split this schema out into separate schemas? Right now, this schema is overloaded between representing *both* bank transfers and mobile transfers. 

We already have a [MobileMoney](https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#7322-mobilemoney) Fiat Account schema; I think you could probably re-use this for mobile payments. Instead of having a user select a machine-readable "mobile provider code", you could have them select a human-readable/parsable string from the `SupportedOperatorEnum` list. 

For bank transfers, it sounds like you could also have two separate schemas: one for bank accounts where IFSC code is required, and another for bank accounts where IBAN is required.

Just to be clear, the reason I'm strongly in favor of having multiple schemas for each different type of account is due to the end-user experience we'll be able to present to the user. If we just have one huge, monolithic schema that includes *all* different account types, it becomes extremely difficult to prompt the user for the correct information. From a client developer's perspective, using one monolithic schema makes it significantly more difficult to implement a form or UI to prompt the user for correct information in a way that is intelligible and makes sense to the end user.

To reiterate; I think my recommendation here would be to:
1. Try to re-use the `MobileMoney` schema for mobile transfers. You can add to the `SupportedOperatorEnum` list for enumerating allowed providers by setting the `allowedValues` field in the quote response. In the backend, you can perform a mapping from the selected provider string to the code that you require.
2. Add a new schema for IFSC bank accounts. You may actually be able to re-use the [AccountNumber](https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#7321-accountnumber) Fiat Account schema, with the stipulation that when `country` is set to `"IN"`, `accountNumber` adheres to IFSC syntax. I would try to re-use this schema if possible.
3. Add a new schema for IBAN accounts. Since IBAN codes are internationally recognizable, IBAN-based fiat account information belongs in its own schema, rather than re-using the `AccountNumber` schema.

fiatconnect-api.md
```diff
@@ -1796,6 +1796,293 @@ is below:
...
+    "accountNumber": `string`,
+    "fiatAccountType": `FiatAccountTypeEnum.BankAccount`
+}
+```
+
+The `country` field should be a [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) country code. The field is for providers to specify which country this data is meant for.
+
+Based on the country `institutionName` and `institutionCode` should be supplied. AllowedValues Enum is mentioned in section 7.3.2.2.1 
+
+In case the type of payment is bank transfer, `institutionCode` and `accountNumber` field will be updated.
```
jophish, line 1817: I don't see an `accountNumber` field in the schema anywhere.

fiatconnect-api.md
```diff
@@ -1796,6 +1796,293 @@ is below:
...
+{
+    "accountName": `string`,
+    "institutionName": `string`,
+    "institutionCode": `string`,
+    "accountNumber": `string`,
+    "fiatAccountType": `FiatAccountTypeEnum.BankAccount`
+}
+```
+
+The `country` field should be a [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) country code. The field is for providers to specify which country this data is meant for.
```
jophish, line 1813: I don't see a `country` field in the schema.

SonalBhanawat: resolved

fiatconnect-api.md
```diff
@@ -1796,6 +1796,897 @@ is below:
...
+`CheckingAccountIban` is a Fiat Account schema that represents accounts for transactions across `EUR` currency.
+
+
+```
+{
+    "accountName": `string`,
+    "institutionName": `string`,
+    "iban": `string`,
+    "accountNumber": `string`,
+    "country": `string`,
```
cajubelt, line 1824: iban actually includes a country code as the prefix

cajubelt: though this field may still be useful so the provider can more easily use "allowedValues" to restrict which countries are allowed

fiatconnect-api.md
```diff
@@ -1796,6 +1796,897 @@ is below:
...
+
+
+```
+{
+    "accountName": `string`,
+    "institutionName": `string`,
+    "bankCode": `string`,
+    "accountNumber": `string`,
+    "country": `string`,
+    "fiatAccountType": "CheckingAccountBankCode"
```
cajubelt, line 1841: doesn't this represent a bank account? if so, the fiatAccountType should be ``FiatAccountTypeEnum.BankAccount``

fiatconnect-api.md
```diff
@@ -1796,6 +1796,897 @@ is below:
...
+    "institutionName": `string`,
+    "ifsc": `string`,
+    "accountNumber": `string`,
+    "country": `string`,
+    "fiatAccountType": "CheckingAccountIfsc"
+}
+```
+####  7.3.2.3. `CheckingAccountIban`
+
+`CheckingAccountIban` is a Fiat Account schema that represents accounts for transactions across `EUR` currency.
```
cajubelt, line 1815: To my knowledge, it's actually an international standard not limited to Europe

fiatconnect-api.md
```diff
@@ -1796,6 +1796,897 @@ is below:
...
+    "iban": `string`,
+    "accountNumber": `string`,
+    "country": `string`,
+    "fiatAccountType": "CheckingAccountIban"
+}
+```
+
+####  7.3.2.4. `CheckingAccountBankCode`
+
+`CheckingAccountBankCode` is a Fiat Account schema that represents accounts for transactions across `NG`, `GH`, `KE`, `ZA`, `PH`, `SN`, `CN`, `GB`, `SG`, `TZ`, `HK` currencies.
```
cajubelt, line 1831: Aren't these countries, not currencies? 

Also, there's many countries which use checking accounts... not sure the list of countries where this fiat account schema is accepted by BitSSA belongs here

fiatconnect-api.md
```diff
@@ -1796,6 +1796,897 @@ is below:
...
+    "bankCode": `string`,
+    "accountNumber": `string`,
+    "country": `string`,
+    "fiatAccountType": "CheckingAccountBankCode"
+}
+```
+
+####  7.3.2.5. `CheckingAccountMobile`
+
+`CheckingAccountMobile` is a Fiat Account schema that represents accounts for transactions across `NG`, `GH`, `KE`, `UG`, `SN`, `BJ`, `ML`, `CI`, `NE`, `TG`, `GW`, `BF`, `SL`, `RW`, `CM`,`CG` currencies.
```
cajubelt, line 1847: There is already a mobile money fiat account schema, I'd prefer if we didn't tack on another one that represents pretty much the same thing. I suggest you add the mobile money providers you want to support-- MTN, TIGO, and Vodafone-- to the SupportedOperatorEnum rather than create a new bespoke FiatAccount schema that represents the same kind of account

fiatconnect-api.md
```diff
@@ -1796,6 +1796,897 @@ is below:
...
+    "country":'string'
+}
+```
+
+
+The `country` field should be a [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) country code. The field is for providers to specify which country this data is meant for.
+
+####  7.3.2.2.1 `SupportedBankMobileProvidersEnum`
+
+For `NG` the list of supported mobile provider codes is as mentioned below -
```
cajubelt, line 1865: This list of all provider codes that are accepted per country shouldn't be in the FiatConnect spec IMO; it looks highly specific to BitSSA. Instead, it should be given to the client in a list of allowedValues in a response to the initial quote request, for which `country` is a parameter. (So BitSSA has the chance to tell the client which providers are accepted dynamically when they request a quote.)

This is actually good for BitSSA, because you can add or remove providers dynamically without having to first undergo a spec change (provided that the name of the provider is already listed in an enum in the specification somewhere)

fiatconnect-api.md
msshahanshah: When providing a quote the rate could differ per country. 
Initially, we had the list of countries in allowed values format and as per @jophish comment, we changed this.

fiatconnect-api.md
msshahanshah: We will update this to a more generic statement for all countries that support IBAN account type format.

fiatconnect-api.md
msshahanshah: @cajubelt yes these are countries and the CheckingAccountBankCode schema is supported in these countries only.

fiatconnect-api.md
msshahanshah: We had initially done it in the proposed method of allowdValues but on the partner call, it was suggested that we separate the schema for each.

fiatconnect-api.md
msshahanshah: We had suggested we provide the list of bank names and bank codes in a getBankCode API call,  On the partner call we were told to include this in the schema.

msshahanshah: We will add the bank and mobile provider code to the response for quote out as allowed vaues.

fiatconnect-api.md
SonalBhanawat: @cajubelt - removed the mobile schema, we will utilise the other schema which is already for mobile money

fiatconnect-api.md
SonalBhanawat: @cajubelt - fixed

fiatconnect-api.md
SonalBhanawat: @cajubelt - updated.

PR #24, opened 2022-04-13
Add `/clock` endpoint for time synchronization

jophish:
Small PR to add a simple `/clock` endpoint dedicated to synchronizing client-server clocks. This PR also updates all values representing dates to use the ISO 8601 standard rather than UNIX time, for the sake of consistency.

Will update the TOC and section numbering once the PR looks good to folks.

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -209,6 +209,7 @@
     + [8.2.2. Idempotency Keys](#822-idempotency-keys)
     + [8.2.3. ISO 3166-1](#823-iso-3166-1)
     + [8.2.4. ISO 3166-2](#824-iso-3166-2)
+    + [8.2.5. ISO 8601](#825-iso-8601)
 
 # 1. Introduction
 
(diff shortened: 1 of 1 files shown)


fiatconnect-api.md
```diff
@@ -385,6 +386,48 @@ will be allowed to initiate a transfer. In the case where a user submits KYC det
...
+In particular, the authentication scheme relies on clock synchronization to choose proper issuance and expiration times for session cookies,
+and the quote endpoints return timestamps representing the expiry of the requested quote, which requires client-server clock synchronization
+in order to be interpreted correctly.
+
+Rather than having bespoke clock synchronization solutions for each of these cases, the FiatConnect specification requires API implementors to
+include a single `/clock` endpoint, which simply returns the server's current timestamp. Servers MUST implement this endpoint. With this information,
+the client can determine the offset between their local time and server time, and synchronize on clocks (to an acceptable approximation).
+The exact method by which clients determine the clock offset is an implementation detail left to the client. (For example, a relatively
+easy and inexpensive method that incoroporates round-trip delay is the approach taken by the
+[Network Time Protocol](https://en.wikipedia.org/wiki/Network_Time_Protocol#Clock_synchronization_algorithm).)
```
cajubelt, line 401: to actually use this algo you'd need more information-- in particular, there's no way to get t1 (the time the request hits the server) from the `/clock` endpoint as currently written.

cajubelt: although I guess you could assume t1 and t2 are both equal to the `time` value on the response body

cajubelt (approved):
Looks good, left one minor comment. Thanks for doing the research.

fiatconnect-api.md
jophish: Yeah, I also thought about this and came to the same assumption, but didn't want to be too prescriptive/go into too much detail here. Realistically t1 and t2 will be extremely close to each other anyways. Presumably some clients won't even incorporate RTT into their offset, so just providing a pointer to this here seems like enough for curious client developers.

PR #27, opened 2022-04-15
fix: quote ids swagger

cajubelt:
Updates to keep openapi spec in sync with markdown spec from changes in https://github.com/fiatconnect/specification/pull/19

added quoteId to quote response schema, and made guaranteedUntil required
Screen Shot 2022-04-15 at 7 39 23 PM

added quoteId to transfer request body
Screen Shot 2022-04-15 at 7 38 35 PM

added InvalidQuote error type
Screen Shot 2022-04-15 at 7 38 52 PM

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -475,7 +475,9 @@ definitions:
             type: "number"
           cryptoAmount:
             type: "number"
-          guaranteedUntil?:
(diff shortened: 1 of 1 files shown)


jophish (approved):
Awesome :slightly_smiling_face:

PR #23, opened 2022-04-12
Add FiatType brazilian Real and cReal CryptoType

rafapolo:
Adding Brazilian currency Types [[reference](https://medium.com/celoorg/celo-launches-the-creal-stablecoin-11da0d560c1c)]

PR #28, opened 2022-04-18
fix: update openapi to be in sync with markdown spec

cajubelt:
Includes updates from the following spec changes:
- add cREAL currency [fiatconnect/specification#23](https://github.com/fiatconnect/specification/pull/23)
- add AccountNumber fiat account type [fiatconnect/specification#16](https://github.com/fiatconnect/specification/pull/16)
- add InvalidParameters error [fiatconnect/specification#14](https://github.com/fiatconnect/specification/pull/14)
- add PersonalDataAndDocuments kyc type [fiatconnect/specification#6](https://github.com/fiatconnect/specification/pull/6)

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -615,6 +615,7 @@ definitions:
       - FiatAmountTooHigh
       - CryptoNotSupported
       - FiatNotSupported
+      - InvalidParameters
   ResourceExists:
     type: "string"
     enum: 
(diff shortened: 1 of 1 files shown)


jophish (approved):
Thanks for this; this LGTM, though I just realized that this swagger spec doesn't actually contain the fiat account or kyc schemas themselves, which means the parameters for some of the endpoints are incorrect. probably should be fixed a followup PR, though.

PR #30, opened 2022-04-20
Update `GET transfer/:transferId` response, add proper webhook payload schemas

jophish:
This PR updates the response from the `GET /transfer/:transferId` to include both the `transferId` and `transferAddress` fields. It also finally adds webhook payload schemas into the specification (which were in the types repo, but not here, for some reason...). 

These changes are somewhat related, in that the transfer status webhook payload should maintain parity with the the `GET /transfer/:transferId` response body, which was previously missing critical information that should have been included in the webhook payload (`transferId` and `transferAddress`, as mentioned above). Including `transferAddress` in the regular `GET /transfer/:transferId` response is useful independently as well, since previously, this information was _only available_ in the initial transfer request response, and then never accessible again!

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -171,6 +171,9 @@
             - [3.3.4.3.3.2.1. `ResourceNotFound`](#3343321--resourcenotfound-)
 - [4. Webhooks](#4-webhooks)
   * [4.1. Webhook Requests](#41-webhook-requests)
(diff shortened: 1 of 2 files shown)


cajubelt (approved):

jophish:
@rafapolo I'm not opposed necessarily, but I believe that should belong in a separate PR; feel free to open one if you like!

PR #34, opened 2022-04-26
Remove redundant information from transfer requests.

jophish:
Since the amount to transfer is already encoded in the `quoteId` used for the transfer, it doesn't make sense to include a duplicate field for `amount` when initiating transfers. Same goes for `cryptoType` and `fiatType`.

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -1463,10 +1463,6 @@ The `POST /transfer/in` endpoint is used to initiate a new transfer in from fiat
 
 The request body must contain the following fields:
 
-* `fiatType`: {`FiatTypeEnum`} [REQUIRED]
-  - The type of fiat currency to transfer in from.
-* `cryptoType`: {`CryptoTypeEnum`} [REQUIRED]
-  - The type of cryptocurrency to transfer in to.
 * `fiatAccountId`: {`string`} [REQUIRED]
   - The fiat account ID to use for the transfer.
 * `quoteId`: {`string`} [REQUIRED]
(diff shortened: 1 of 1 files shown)


fiatconnect-api.md
```diff
@@ -1544,8 +1538,8 @@ If a transfer is not allowed for a generic reason (such as unacceptable transfer
 
 ###### 3.3.4.1.3.2.3. `InvalidQuote`
 
-If the quote associated with `quoteId` is expired, or if no quote is found for the user with a matching `quoteId`, the server MUST
-reject the transfer and return an `InvalidQuote` error. 
+If the quote associated with `quoteId` is expired, or if no quote is found for the user with a matching `quoteId`, or if the provided quote is for the
```
cajubelt, line 1541: nit: might make sense to update the "semantics" section to indicate that not only the exchange rate and fees, but also the amount, cryptoType, and fiatType, should be used from the original quote

fiatconnect-api.md
```diff
@@ -1652,8 +1640,8 @@ If a transfer is not allowed for a generic reason (such as unacceptable transfer
 
 ###### 3.3.4.2.3.2.3. `InvalidQuote`
 
-If the quote associated with `quoteId` is expired, or if no quote is found for the user with a matching `quoteId`, the server MUST 
-reject the transfer and return an `InvalidQuote` error.
+If the quote associated with `quoteId` is expired, or if no quote is found for the user with a matching `quoteId`, or if the provided quote is for the
```
cajubelt, line 1643: same here

fiatconnect-api.md
cajubelt: relatedly, this sentence from `3.3.1.2.3.1` may need to be updated:
```
The `quote.quoteId` field
is a globally unique identifier for the quote, and is used by the client to initiate a transfer at the
conversion rate and fee amount given by the quote.
```

cajubelt (commented):
Seems reasonable, and I imagine it will reduce the number of error cases to think about on the providers' side. Though I noted a couple places we could adjust for clarity.

Also, I'd like to make sure providers know to expect this and are ok with it.

jh2oman (approved):
Looks good! great simplification

cajubelt (approved):

PR #35, opened 2022-04-26
fix: auth, clock openapi

cajubelt:
Looks like the new SIWE auth scheme and the clock endpoint are missing from the openapi spec. This adds them

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -4,6 +4,10 @@ info:
   version: "1.0.0"
   title: "FiatConnect Specification"
 tags:
+- name: "clock"
+  description: "Sync clocks between server and client"
+- name: "auth"
+  description: "Authentication"
 - name: "quote"
   description: "Requests quotes for transfers with the user's desired parameters"
 - name: "kyc"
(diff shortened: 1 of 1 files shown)


swagger.yaml
```diff
@@ -16,6 +20,47 @@ schemes:
...
+        - in: "body"
+          name: "signature"
+          description: "An EIP-191 (for EOA) or EIP-1271 (for contract-owned accounts) signature for the SIWE message"
+          required: true
+          type: "string"
+      responses:
+        "200":
+          description: "successful operation"
+          schema:
+            type: "string"
```
cajubelt, line 59: I *think* the return type is just the session cookie as a string, but correct me if I'm wrong

cajubelt: https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#32152-responses

satish-ravi: The body would just be empty. The session cookie will be part of the `Set-Cookie` response header.

cajubelt: ah

swagger.yaml
```diff
@@ -16,6 +20,45 @@ schemes:
...
+            $ref: "#/definitions/ClockResponse"
+  /auth/login:
+    post:
+      summary: "Log in with a SIWE message and get a session cookie"
+      tags:
+        - "auth"
+      consumes:
+        - "application/json"
+      produces:
+        - "application/json"
```
satish-ravi, line 43: Remove? Since there's no response body

satish-ravi (approved):

swagger.yaml
cajubelt: I think there is if an error occurs

satish-ravi: 🤦‍♂️ 😆

PR #40, opened 2022-05-10
Swagger spec enums and equivalent paths

MuckT:
### Description

- Fixes enums by moving definitions above paths and using [Reusable Enums](https://swagger.io/docs/specification/2-0/enums/) - Enums previously used are only valid in the 3.0 spec.

- Fixes equivalent paths between `/accounts/{fiatAccountSchema}` and `/accounts/{fiatAccountId}` by changing the later to `/accounts/delete/{fiatAccountId}`. This requires a code change in the spec - more details about this type of issue can be seen in https://github.com/swagger-api/swagger-editor/issues/1677#issuecomment-365826052

Validated with [Swagger Editor](https://editor.swagger.io/).


swagger.yaml
```diff
@@ -294,7 +558,7 @@ paths:
                   $ref: '#/definitions/FiatAccountInfoResponse'
       security:
         - siwe_auth: []
-  /accounts/{fiatAccountId}:
+  /accounts/delete/{fiatAccountId}:
```
silasbw, line 561: Wondering if this ambiguity warrants further discussion around `POST /accounts/{fiatAccountSchema}`.  I would actually expect that endpoint to take all parameters in the body (vs some in the body and some in the path).  Is there a reason to use both path and body for that POST endpoint @jophish @jh2oman @cajubelt ?

jh2oman (approved):

MuckT:
@jh2oman `fiatconnect-api.md` updated in https://github.com/fiatconnect/specification/pull/40/commits/d8a567e1f2e2c4c5cbf7c8cd88ac01daa4e174ca!

swagger.yaml
cajubelt: all in the body does seem a bit cleaner

satish-ravi: +1 to having `fiatAccountSchema` in the body and just doing `/accounts/{fiatAccountId}` here

lisaqyli: +1 all in the body feels cleaner to me as well.  I don't think we have `fiatAccountId` on the request to add new fiat account

fiatconnect-api.md
```diff
@@ -141,7 +141,7 @@
           - [3.4.3.2.2.1. HTTP `200`](#343221-http-200)
         - [3.4.3.2.3. Semantics](#34323-semantics)
           - [3.4.3.2.3.1. Success](#343231-success)
-      - [3.4.3.3. `DELETE /accounts/:fiatAccountId`](#3433-delete-accountsfiataccountid)
+      - [3.4.3.3. `DELETE /accounts/delete/:fiatAccountId`](#3433-delete-accountsfiataccountid)
```
lisaqyli, line 144: I'm not sure about what's the benefit of having `delete` in the path?

MuckT: Swagger cannot differentiate between `/accounts/{fiatAccountId}` and `/accounts/{fiatAccountSchema}` as they are equivalent paths.

swagger.yaml
jophish: Yeah, I believe it would just be `POST /accounts` with the request body being e.g:

```
{
  fiatAccountSchema: ...
  data: {
   ...
  }
}
```

or, something like that.

jh2oman (approved):

PR #49, opened 2022-05-20
update BankAccount to AccountNumber

lisaqyli:
Shall we update BankAccount to AccountNumber for FiatAccountTypeEnum? 
is FiatAccountTypeEnum vs FiatAccountSchema one to one?  will think more next week

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -2072,7 +2072,7 @@ represents what *kind* of account that schema represents.
 
 ```
 [
-	`BankAccount`,
+	`AccountNumber`,
 	`MobileMoney`,
 	`DuniaWallet`
 ]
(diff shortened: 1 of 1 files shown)


jophish:
@lisaqyli There's a one-to-many mapping from values of `FiatAccountType` to `FiatAccountSchema`, and each `FiatAccountSchema` is mapped to exactly once; in other words, there could be many different schemas for each `FiatAccountType`. The `BankAccount` account type is meant to represent... a bank account! The way I think about the `AccountNumber` schema right now is as a "catchall" for accounts that merely require an account number, with syntactical restrictions on the `accountNumber` field based on locale. In practice, `AccountNumber` is only used for Nigerian accounts at the moment, and it's unclear if it will actually be expanded to be used for other regions. I would actually be in support of changing `AccountNumber` to something like `NGBankAccount` or something of the sort, and keeping that schema specific to accounts used only in Nigeria.

PR #54, opened 2022-05-31
Add specific requirements for webhook retries

jophish:
This came from a discussion @satish-ravi and I had following a comment from a partner in the Discord channel. This PR adds a requirement that webhooks must retry 4 times after an initial failure, with a recommendation on exponential retry times.


fiatconnect-api.md
```diff
@@ -1820,8 +1820,9 @@ different providers. The `eventId` field is a unique event ID generated by the s
 ensure that events are not processed more than once. The `timestamp` field represents the exact time this event was initially generated, represented
 as time since Epoch in seconds. The `payload` field contains the actual event data, and will vary depending on the selected `eventType`.
 
-Servers SHOULD implement retries with exponential backoff when sending payloads to a client's webhook. Once a server has received an HTTP `200`
-status code from the client's webhook, it MUST stop retrying.
+Servers MUST implement retries with exponential backoff when sending payloads to a client's webhook. Servers MUST retry at least 4 times if the original
+request does not return an HTTP `200` status code, but MAY retry more. The waiting period between the first 4 retries SHOULD be roughly 3, 66, 731, 4098 seconds respectively,
```
cajubelt, line 1824: doesn't exponential backoff usually mean waiting by some period that increases by a constant factor each time? where do these numbers come from?

cajubelt (approved):

fiatconnect-api.md
satish-ravi: I think these were taken from [persona](https://docs.withpersona.com/docs/webhooks#retry-logic), looks like these follow `n^6+2`. Maybe we can suggest a simpler sequence?

satish-ravi (approved):

lisaqyli (approved):

PR #56, opened 2022-06-04
Document a versioning scheme and the pre 1.0.0 state

silasbw:


silasbw:
Consider this a proposal!

jophish (commented):
@silasbw Thanks a ton for this; this is something that's been on my mind more and more recently. With partners coming to us for final review, we really need to start thinking about locking at some version. As soon as a partner gets the green light on their implementation and a wallet/dapp starts consuming their API, changes to the spec become a lot harder to make. In order to allow for non-backwards compatible changes, APIs will probably have to broadcast the version of FiatConnect they support somehow (over an endpoint, in a header, etc.). The SDK will also need to be pinned to a spec version, and probably check that any URL passed to it supports the same major version.

Major version bumps (e.g. v1 -> v2) will probably have to be done with the following process:
1. Partner deploys v2-compliant API at e.g. `provider.com/v2/` and makes address well-known
2. New major version of FiatConnect SDK released that only supports v2-compliant APIs
3. Clients update to the newest major FiatConnect SDK version, replacing the old `provider.com/v1/` URL with `provider.com/v2`
4. Once all consumers of the old `provider.com/v1/` have migrated to the new URL, the provider can shut down the v1 implementation.

All this to say... it might be worth considering how APIs can share the version of the FiatConnect spec that they implement with the client, and adding it to the spec as part of this PR.

silasbw:
That makes sense.  For communicating the version of the spec to client, I often see a `GET /openapi` or `GET /swaggerapi` endpoint.  That might be a bit cumbersome at this point (because it would require merging several versions of the API spec into a single specification).  One potential step in that direction is something like `GET /v2/openapi` or even `GET /v2` (which would return 200 if v2 is supported).  A client could start by calling `GET /v${HIGHEST_VERSION_THEY_SUPPORT` and go down from there to find the highest supported version.

For this PR, how about it:

1. explicitly documents adding versioning to the path; and
2. adds the `GET /v1` endpoint to the current specification that's approaching v1.

Do those sound like good next steps for this PR?

cajubelt (approved):

jophish (approved):

PR #61, opened 2022-06-15
fix: clarify nonce should be at least 8 characters

cajubelt:
The SIWE docs make it more clear that the nonce should be *at least* 8 characters. This led to a misunderstanding with at least one provider API so far. https://docs.login.xyz/general-information/siwe-overview/eip-4361#message-field-descriptions


satish-ravi (approved):

jophish (approved):
TY for catching this!

Issue #42, opened 2022-05-10
[FiatType] Include Latin American fiat currencies


andreujuanc:
Hi, 
We noticed that lots of the local fiat currencies in Latin America are missing from the type https://github.com/fiatconnect/specification/blob/main/swagger.yaml#L463

Is it possible to do a PR to add them? Will it be taken into account? 
We'll reach out to the Valora team via email about this as well.

Thanks

rafapolo:
I'm glad to help out this one, since I've added the Brazilian one.

| Country | Currency | ISO-4217 |
| --- | --- | --- |
| Argentina | Argentine peso | ARS |
| Bolivia | Bolivian boliviano | BOB |
| Brazil | Brazilian real | BRL |
| Chile | Chilean peso | CLP |
| Colombia | Colombian peso | COP |
| Ecuador | United States dollar | USD |
| Falkland Islands (UK) | Falkland Islands pound | FKP |
| French Guiana (France) | European euro | EUR |
| Guyana | Guyanese dollar | GYD |
| Paraguay | Paraguayan guarani | PYG |
| Peru | Peruvian sol | PEN |
| South Georgia Island (UK) | Pound sterling | GBP |
| Suriname | Surinamese dollar | SRD |
| Uruguay | Uruguayan peso | UYU |
| Venezuela | Venezuelan bolivar | VES |

Just not sure if we should add ALL countries currencies at once, or if the CICO Providers would be the ones announcing the FIATs they support; otherwise we would have listed FIATs without Providers.

andreujuanc:
>  if the CICO Providers would be the ones announcing the FIATs

This could potentially be a breaking change in the spec. But I love this idea.

> Just not sure if we should add ALL countries currencies at once

We can just compile and add whatever we all need. This can be the easiest and least disruptive.

These are currencies that we support and are not in your list:

<html><body>


Country | Currency | ISO-4217
-- | -- | --
Mexico| Mexican peso | MXN
Panama| Panamanian balboa| PAB


</body>
</html>

PR #63, opened 2022-06-16
chore(currencies): added latam fiat currencies

gusparr28:


gusparr28:
closes https://github.com/fiatconnect/specification/issues/42

missing Latin American fiat currencies were added to FiatType https://github.com/fiatconnect/specification/blob/main/swagger.yaml#L39

swagger.yaml
```diff
@@ -55,6 +55,19 @@ definitions:
       - RWF
       - CNY
       - XAF
+      - ARS
```
cajubelt, line 58: thanks, can you also add these to the FiatTypeEnum in the `fiatconnect-api.md` file?

gusparr28: thanks for the review. Sure, already done

cajubelt (approved):
thanks!

PR #68, opened 2022-07-15
refactor(amounts)!: changed to float to string

cajubelt:
Using floats for amounts could result in errors due to floating point rounding errors, see [here](https://floating-point-gui.de/) for more info

BREAKING CHANGE: all float type inputs/outputs must changed to string types


fiatconnect-api.md
```diff
@@ -869,7 +869,7 @@ about the corresponding fiat account type. Each object MUST contain a `fiatAccou
 optional `allowedValues` field. The `allowedValues` object is an optional mapping from any number of keys in the selected fiat account schema to values that are allowed for that key.
 This is identical in purpose and function to the `allowedValues` field for KYC schemas, discussed earlier.
 
-The `fiatAccount[FiatAccountTypeEnum].fee` field is an optional return value, used to represent an optional fixed fee for the transfer
+The `fiatAccount[FiatAccountTypeEnum].fee` field is an optional return value, used to represent an optional fixed fee amount for the transfer
```
jh2oman, line 872:
```suggestion
The `fiatAccount[FiatAccountTypeEnum].fee` field is an optional return value, used to represent an optional fixed fee as a string-ified numerical amount (e.g. `"1.0"`) for the transfer
```

fiatconnect-api.md
```diff
@@ -1042,7 +1042,7 @@ about the corresponding fiat account type. Each object MUST contain a `fiatAccou
...
```
jh2oman, line 1045:
```suggestion
The `fiatAccount[FiatAccountTypeEnum].fee` field is an optional return value, used to represent an optional fixed fee as a string-ified numerical amount (e.g. `"1.0"`) for the transfer
```

jh2oman (approved):
Just some suggestions for consistency. Looks good though 👍

satish-ravi (approved):

fiatconnect-api.md
```diff
@@ -1890,51 +1894,88 @@ the UNIX timestamp that the request was sent. The second will be of the form `s=
...
+In the interest of preventing rounding errors, [floating point errors](https://floating-point-gui.de/), or similar errors, numerical amounts 
+MUST be communicated as strings for every request parameter and every response field throughout the FiatConnect API. 
+
+Clients and servers alike SHOULD NOT cast fiat or crypto amounts to floating point numbers or integers, but should instead
+consider libraries such as [bignumber.js](https://github.com/MikeMcl/bignumber.js/) for handling large numbers and decimals 
+to an arbitrary number of decimal places.
+
+## 6.1 Units
+
+Except where otherwise noted, amounts of cryptocurrencies MUST be given in the units of that currency, e.g. `1.5` Celo means
```
jophish, line 1908: Perhaps add a note here specifying that the precision given for cryptocurrency amounts MUST NOT exceed the maximum possible precision for the given currency (i.e., that specified by its value for ERC20's [decimals](https://docs.openzeppelin.com/contracts/3.x/erc20#a-note-on-decimals))

fiatconnect-api.md
```diff
@@ -1890,51 +1894,88 @@ the UNIX timestamp that the request was sent. The second will be of the form `s=
...
+## 6.2 Decimal Places
+
+Given that amounts are communicated in divisible units-- such as US Dollars, which are divisible into `100` cents, or Celo, which 
+is divisible into `10^18` wei--  it is expected that the client or server will sometimes need to communicate non-integer amounts. 
+For all amounts, [decimal notation](https://en.wikipedia.org/wiki/Decimal#Decimal_notation) MUST be used-- for example, 
+One half should be given as `0.5`.
+
+## 6.3 Precision
+
+FiatConnect-compliant API's MUST observe 18 decimal places of precision for all cryptocurrencies and 2 decimal places of
```
jophish, line 1923: Ah, you call out precision specifically here. Maybe a bit nit-picky, but ERC20 tokens don't necessarily have 18 decimals of precision; it's less common, but I've run into it plenty. No reason not support those tokens in the future IMO.

fiatconnect-api.md
```diff
@@ -1890,51 +1894,88 @@ the UNIX timestamp that the request was sent. The second will be of the form `s=
...
+precision for all fiat currencies. 
+
+Note that this does NOT mean a CICO provider is required to support transfers of 
+extremely small amounts; see [here](#3411322-cryptoamounttoolow).
+
+Note also that while client and server MUST give amounts in the full degree of precision to unambiguously communicate
```
jophish, line 1929: Huh, it sounds like what you're saying is that clients/servers _don't_ need to use the maximum precision specified by the currency, just enough to disambiguate the amount from other amounts... which is practically to say, "no lower bound on precision in amounts, only upper bounds" right? I think this could maybe be phrased to be a bit more straightforward.

fiatconnect-api.md
```diff
@@ -1890,51 +1894,88 @@ the UNIX timestamp that the request was sent. The second will be of the form `s=
...
+one and a half Celo (not `1.5e-18` Celo, which would be the case if amounts were instead given in 
+[wei](https://www.investopedia.com/terms/w/wei.asp#:~:text=What%20Is%20a%20Wei%3F,one%20quintillionth%20of%20an%20ether.)).
+
+Amounts of fiat currencies MUST be given in the units of that currency, e.g. `5.00` USD means five USD.
+
+## 6.2 Decimal Places
+
+Given that amounts are communicated in divisible units-- such as US Dollars, which are divisible into `100` cents, or Celo, which 
+is divisible into `10^18` wei--  it is expected that the client or server will sometimes need to communicate non-integer amounts. 
+For all amounts, [decimal notation](https://en.wikipedia.org/wiki/Decimal#Decimal_notation) MUST be used-- for example,
```
jophish, line 1918: nit: The [decimal notation](https://en.wikipedia.org/wiki/Decimal#Decimal_notation) doesn't actually mandate the separator used to be a literal dot (`.`); it only mentions that dots are often used, but commas a used as well. Probably good to explicitly give the syntax. A regex like `/^([0-9]+\.?[0-9]*|\.[0-9]+)$/` could work.

jophish (commented):
Looks good! A couple of comments to make things a bit more explicit/clear

fiatconnect-api.md
cajubelt: I'd actually argue that when it comes to decimal places, there is no upper bound, but there is a lower bound

cajubelt: though the server can ignore anything smaller than 1 wei (or whatever atomic unit applies for that token/currency)

cajubelt: i can try to rephrase, though I'll avoid the upper/lower bound thing since we're already reading that 2 different ways apparently

fiatconnect-api.md
cajubelt: OOPS

fiatconnect-api.md
cajubelt: updated

cajubelt: didn't realize that about erc20

fiatconnect-api.md
cajubelt: I'm trying out a "significant figures" and "trailing zeroes" explanation instead

fiatconnect-api.md
cajubelt: ended up removing the reference to decimal notation, too loosey goosey

jh2oman (approved):

fiatconnect-api.md
```diff
@@ -1915,21 +1915,22 @@ Amounts of fiat currencies MUST be given in the units of that currency, e.g. `5.
...
+separator MUST be the period symbol: `.` . For example, one half should be given as `0.5`. 
+
+Put another way, all amounts MUST match the following regex: `/^[0-9]+\.?[0-9]*$/` .
 
 ## 6.3 Precision
 
-FiatConnect-compliant API's MUST observe 18 decimal places of precision for all cryptocurrencies and 2 decimal places of
-precision for all fiat currencies. 
+FiatConnect-compliant API's MUST observe the maximum number of decimal places of precision for every cryptocurrency in `CryptoTypeEnum`, and 2 decimal places of
+precision for all fiat currencies. (Most, but not all, ERC-20 tokens offer up to 18 decimal places of precision.)
```
jophish, line 1926: super-nit: ERC20 tokens can also specify [more than 18 decimals](https://ethereum.stackexchange.com/questions/118896/can-an-erc-20-have-more-than-18-decimals#:~:text=Theoretically%2C%20you%20could%20use%20any,tokens%20in%20terms%20of%20supply. of precision

jophish (approved):
Just one tiny comment, but LGTM!

fiatconnect-api.md
cajubelt: it's still true that most offer up to 18 decimals

cajubelt: for example, the Celo token offers up to 18 decimal places of precision

PR #70, opened 2022-07-26
feat(quotes): Add ability to preview quotes

jophish:
We've gotten feedback from providers that it may be helpful to be able to distinguish between quotes whose intent are to simply get a preview, vs. actually performing a transfer. The pain here is that currently, it's impossible to determine if a quote has a high likelihood of being executed, which means that certain provider implementations that rely on "locking up" liquidity for the duration of a quote's life may needlessly lock up large amounts of their liquidity for clients with no intention of actually using their services.

This PR adds an optional `preview` parameter to the quote endpoints. A new pattern for clients who integrate with multiple providers might be:
* Get preview quotes from all providers
* Let the user select a provider based off the preview quotes
* Once the provider is selected and account information is linked, get a real quote form that provider
* Execute a transfer using the real quote we just received

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -971,7 +971,7 @@ On success, the server MUST return an HTTP `200`, with the following response bo
 		cryptoType: `CryptoTypeEnum`,
 		fiatAmount: `string`,
 		cryptoAmount: `string`,
-		quoteId: `string`,
+		quoteId?: `string`,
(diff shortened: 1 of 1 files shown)


jh2oman (approved):
Problem makes sense and solution looks good.

Can you create a ticket for us to update the wallet once this feature is pushed through? I think it'll be a slightly larger update to utilize than most other spec changes.

PR #72, opened 2022-07-26
feat(fiatAccount): Add fiatAccountSchema to the fiatAccount response

jh2oman:
This was likely an oversight on our part 🤦


jophish (approved):
Can't believe we missed this one! :facepalm: Thanks for the fix. We should share this with partners before merging.

PR #76, opened 2022-08-08
fix(swagger): Update account field naming, fix GetFiatAccountsResponse validation

jophish:
Two small things in this PR:

1. The field names in account data weren't matching the spec; this updates them to match e.g. [here](https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#343221-http-200).
2. The syntax used to specify the key type for the `GET /accounts` response (`<FiatAccountTypeEnum>`) was not working correctly during validation, so I've updated the swagger spec to use `additionalProperties` instead and drop this constraint.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -321,7 +321,7 @@ definitions:
         $ref: '#/definitions/CryptoType'
       amountProvided:
         type: "string"
-      amountRecieved:
+      amountReceived:
         type: "string"
       fee:
         type: "string"
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -239,16 +239,16 @@ definitions:
     type: "object"
     required:
       - fiatAccountId
-      - name
-      - institution
+      - accountName
+      - institutionName
       - fiatAccountType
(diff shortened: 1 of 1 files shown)


anthonydaoud (approved):

swagger.yaml
```diff
@@ -239,29 +239,28 @@ definitions:
...
+      institutionName:
         type: "string"
       fiatAccountType:
         $ref: '#/definitions/FiatAccountTypeEnum'
       fiatAccountSchema:
         $ref: '#/definitions/FiatAccountSchemaEnum'
   GetFiatAccountsResponse:
     type: object
-    properties:
-      <FiatAccountTypeEnum>:
```
satish-ravi, line 260: this not working is a limitation of the validator package we use?

satish-ravi (approved):

swagger.yaml
jophish: Yeah, that's correct; this is a workaround.

PR #81, opened 2022-09-09
docs(fee): update swagger description to match fees

XrayBravoGolf:
This addes to change in documentation in #78 by moving the `QuoteResponse` fee entries to under `quote` from nested under `fiatAccount`.


cajubelt (approved):
thank you!

PR #86, opened 2022-11-18
Remove `allowedValues` from KYC schemas in quote response

jophish:
At present, the utility of the `allowedValues` field for KYC schemas is not immediately clear. Among current FiatConnect API providers, the feature is unused, and it places a burden on client client implementors to support it for no obvious benefit. To be clear, there is a distinct difference in actual usefulness between `allowedValues` for KYC schemas vs. Fiat Account schemas. The semantics of `allowedValues` intend to allow providers to give users a list of _options_ to select from for a particular field. For Fiat Account schemas, this is useful (e.g., to allow users to select between several different institutions), but for KYC information, the information provided by users is generally likely to be unique across all users, calling into question the function of the `allowedValues` field for KYC schemas.

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -800,12 +800,7 @@ On success, the server MUST return an HTTP `200`, with the following response bo
 	},
 	kyc: {
 		kycRequired: `boolean`,
-		kycSchemas: {
-			kycSchema: `KycSchemaEnum`,
-			allowedValues: {
-				[string]: `string[]`
-			}
-		}[]
+		kycSchemas: `KycSchemaEnum`[]
 	},
 	fiatAccount: {
 		[FiatAccountTypeEnum]: {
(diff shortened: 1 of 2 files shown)


fiatconnect-api.md
```diff
@@ -800,12 +800,7 @@ On success, the server MUST return an HTTP `200`, with the following response bo
 	},
 	kyc: {
 		kycRequired: `boolean`,
-		kycSchemas: {
-			kycSchema: `KycSchemaEnum`,
-			allowedValues: {
-				[string]: `string[]`
-			}
-		}[]
```
cajubelt, line 808: I wonder if it would be worth just eliminating the "allowedValues" field for now, and migrating from a list of objects to a list of KYC schemas later on, to limit the scope of the changes that Valora and CICO providers will need to make in the near term

fiatconnect-api.md
```diff
@@ -891,7 +885,8 @@ Each `fiatAccount[FiatAccountTypeEnum]` MUST correspond to a fiat account type t
 `fiatAccount[FiatAccountTypeEnum].fiatAccountSchemas` is a list of objects, where each object represents a fiat account schema that can be used to communicate data
 about the corresponding fiat account type. Each object MUST contain a `fiatAccountSchema` field representing the actual fiat account schema that can be used, and an
 optional `allowedValues` field. The `allowedValues` object is an optional mapping from any number of keys in the selected fiat account schema to values that are allowed for that key.
-This is identical in purpose and function to the `allowedValues` field for KYC schemas, discussed earlier.
+For example, if a server wants to limit the selection for certain fields on the schema, `allowedValues` can contain lists of selectable values for those fields.
+On the client-side, this could be used to render a list of options, for example.
```
cajubelt, line 889:
```suggestion
If a server wants to limit the selection for certain fields on the schema, `allowedValues` can contain lists of selectable values for those fields.
On the client-side, this could be used to render a list of options.
```
Nit: these just seem like details to me, not examples

cajubelt (commented):
I suggest we either split the change into 2 repos as explained in my comment, or merge this to a dev branch that will become v2 at some point down the road

fiatconnect-api.md
satish-ravi: I think even dropping `allowedValues` will be a breaking change since a required field will no longer be present.

jophish:
@cajubelt Given that Alpha Fortress has a use case for this, and that this would likely introduce a breaking change, I like shelving this in a v2 dev branch for the time being; that seems like a good place to keep breaking, non-urgent changes until we're ready to release v2.

PR #88, opened 2022-12-21
refactor(mobile-money): change mobile money operator field to string

cajubelt:
There are only 4 supported operators in the spec, and about 144 in sub-saharan Africa alone. Given that this list may also change over time, I don't think it's wise to try to maintain an enum so large in the FiatConnect spec and all the downstream repos that depend on it (particularly https://github.com/fiatconnect/fiatconnect-types).

I suggest that CICO providers use allowedValues to enumerate the mobile money operators supported by their API. Otherwise, clients are welcome to implement their own drop-downs or auto-complete algorithms for this field to try to proactively sanitize the user inputs and prevent users from getting transfers rejected on account of bad `operator` inputs. (Again, CICO providers are ultimately responsible for advertising the list of mobile money providers that they will support, which they can do using the allowedValues field on a quote response.)

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -236,7 +236,6 @@
     - [9.3.2. Fiat Account Schemas](#932-fiat-account-schemas)
       - [9.3.2.1. `AccountNumber`](#9321-accountnumber)
       - [9.3.2.2. `MobileMoney`](#9322-mobilemoney)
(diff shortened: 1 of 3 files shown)


satish-ravi (approved):

cajubelt:
Got positive feedback from both Bitmama and Bitssa about this change, none on Discord so far.

PR #91, opened 2023-01-24
fix: Make error and transfer enum values consistent with specification

jophish:
To match https://github.com/fiatconnect/fiatconnect-types/pull/117. These were missing from the enum in the spec, which is likely why they were also missing from the types repo.


cajubelt (commented):
need to update the swagger enums as well.

also, worth bumping the version to 1.0.2?

cajubelt (approved):

PR #95, opened 2023-02-23
feat(txhash): Add `txHash` to transfer in status responses

jophish:
Currently, it is difficult, if not impossible, for a client to determine _which_ transaction on the blockchain represents the payment from provider to end-user for a transfer in. While the provider _does_ need to inform the client of the provider-owned address that funds will be sent from, this is not necessarily sufficient to determine the actual transfer that contains the sent funds (and even when it is "sufficient", it can be challenging and error-prone to track down the transaction). 

For example, if a user initiates two "transfer in" transactions _for similar amounts_ in parallel with a provider, and the provider sends funds from the same address for both transfers, it would be extremely difficult to determine which blockchain TX corresponds to which FC transfer.

This PR updates both the transfer status endpoint, and the `WebhookTransferInStatusEventSchema` webhook payload to contain an optional `txHash` field, so that clients can associate transfers in with their corresponding blockchain TX.

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -1731,7 +1731,8 @@ On success, the server MUST return an HTTP `200` status code, along with a respo
 	fee?: `string`,
 	fiatAccountId: `string`,
 	transferId: `string`,
-	transferAddress: `string`
+	transferAddress: `string`,
+	txHash?: `string`
 }
 ```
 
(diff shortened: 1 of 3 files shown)


jh2oman (approved):

cajubelt (commented):
remember to update the spec version for this

PR #96, opened 2023-03-03
feat(fiat-type): add ZMW

cajubelt:
to support transfers in Zambia https://en.wikipedia.org/wiki/Zambian_kwacha

per request [here](https://discord.com/channels/1020413508350906470/1020413509172994160/1080607737479700490)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "specification",
-  "version": "1.0.3",
+  "version": "1.1.0",
   "description": "FiatConnect API Specification",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "specification",
-  "version": "1.0.3",
+  "version": "1.1.0",
```
jophish, line 3: Does this bump a whole minor version? Why not patch? (Or should the previous changes themselves also have been minor bumps?)

jophish (approved):
A, with question about versioning.

Issue #93, opened 2023-02-15
User Initiated Transactions Proposal


orlax:
In the case of Latin American Countries, it is not possible to withdraw or debit money from an account using the ACH banking APIs. This means that for Transfers In we need the ability to have "User Initiated transactions" where users actively send the money from their bank accounts to the CICO provider. This is normally handled by country wide systems. For example in the case of Colombia it is PSE https://www.pse.com.co/persona

So, for a full description of the problem lets go to this document: https://www.notion.so/orlagames/User-Initiated-Transactions-and-Cash-Payments-proposal-03f8234f4fc14e749a9cb047e1557cc5?pvs=4

once you have read that, here is our current solution: 

- On the transfer in endpoint, things remain the same, we receive the pre-registered Fiat account ID and quote ID: 
`{
	"FiatAccountId":"987002",
   "QuoteID":"acb34623"
}`
- Next CICO providers that need to use this user initiated transactions would use the account information and user information obtained in the KYC process to create the payment intention in the country banking payment provider. 
- on the response object we would return the URL were payment needs to be completed: 
```
{
	transferId: `string`,
	transferStatus: `WaitingForPayment`,
	transferAddress: `CELO Address`, 
	transferDetails: '{method:"PSE", url:"some_url"}'
}
```
By introducing that new transferDetails field, the intention is to have some schemas where we can provide the URLs and the system name being used to procesate payment. Then users can go to that url and complete payment on their own, to mark the transaction as user initiated we can use a new "transferStatus" that states we are waiting for the user to complete the transaction. In this case the clients can check for this status and redirect users to the proper URL so that they can complete payment. 

In this video I talk about the problem and do a mini demo of the flow: 
https://drive.google.com/file/d/1AcMmx5r6dKDnNRWinEBxr6nrNo6mQLeS/view?usp=sharing

so, all in all I think if we agree on the new status and what that transferDetails would look like this could be a minor adition to the spec. 

Let's discuss!

0xtuytuy:
This is a good start @orlax I think however this is quite specific to LATAM.

If we wanted to make his more generic to cover the rest of the world where user-initiated payments are important I would do the following modifications.

I would keep the `transferDetails` key in the response of the `POST` `transfer/in` endpoint as you suggest, but I would take the `method` key that you are proposing out of the object, this would therefore allow the devs to see without going too much down the tree what schema they can expect in the `transferDetails` key
```
{
	transferId: `string`,
[...]

PR #97, opened 2023-03-08
feat(transfers): Add support for user-initiated transfers in

jophish:
Formalizes the discussion present in #93.

This PR adds formal support to the FiatConnect specification for user-initiated transfers in.

At a high level, this is accomplished by introducing two new (capital C) _Concepts_ to the FiatConnect specification: **User Action Type** and **User Action Details**. 

At the time of getting a quote from `POST /quote/in`, if a particular Fiat Account schema will require user action to send fiat funds to the provider, a _User Action Type_ should be specified alongside that schema in the quote response. When initiating the transfer by calling `POST /transfer/in`, if the fiat account being used for that transfer has a schema that was designated as requiring user action by the quote, the `POST /transfer/in` response will contain _User Action Details_, containing information required by the user in order to complete the action needed to progress the transfer. Each _User Action Type_ has a corresponding _User Action Details_ schema, each with different semantics.

I've included initial User Action Type/User Action Details support for `PIX`, `IBAN`, and `PSE`. Let me know if I'm missing something that you'll _immediately_ need support for; I don't want to bloat this PR with support for User Action Types/User Action Details that won't be immediately needed; those can always be added easily in a follow-on PR.

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -230,6 +230,8 @@
     - [9.2.5. `FeeTypeEnum`](#925-feetypeenum)
     - [9.2.6. `FeeFrequencyEnum`](#926-feefrequencyenum)
     - [9.2.7. `FiatAccountSchemaEnum`](#927-fiataccountschemaenum)
+    - [9.2.8. `TransferInUserActionTypeEnum`](#928-transferinuseractiontypeenum)
+    - [9.2.9. `TransferInUserActionDetailsEnum`](#929-transferinuseractiondetailsenum)
   - [9.3. Initial Entity Support](#93-initial-entity-support)
     - [9.3.1. KYC Schemas](#931-kyc-schemas)
       - [9.3.1.1. `PersonalDataAndDocuments`](#9311-personaldataanddocuments)
@@ -241,6 +243,10 @@
       - [9.3.2.3. `DuniaWallet`](#9323-duniawallet)
       - [9.3.2.4. `IBANNumber`](#9324-ibannumber)
       - [9.3.2.5. `IFSCAccount`](#9325-ifscaccount)
+    - [9.3.3. User Action Details Schemas](#933-user-action-details-schemas)
+      - [9.3.3.1. `PIXUserActionDetailsSchema`](#9331-pixuseractiondetailsschema)
+      - [9.3.3.2. `IBANUserActionDetailsSchema`](#9332-ibanuseractiondetailsschema)
+      - [9.3.3.3. `PSEUserActionDetailsSchema`](#9333-pseuseractiondetailsschema)
 - [10. References](#10-references)
   - [10.1. Normative References](#101-normative-references)
     - [10.1.1. [RFC2119]](#1011-rfc2119)
@@ -742,8 +748,8 @@ We assume that the lifecycle of a transfer begins with an end-user requesting a 
 a user providing their desired transfer parameters (crypto type, fiat type, amount), and the client then proceeding to show
 them a list of supported providers.
 
-We assume that different quotes represent transfers that may have differing requirements regarding KYC and fiat accounts. To support this, the quote
-endpoints must return information about KYC and fiat account schemas that are required in order to actually perform a transaction for the requested quote.
+We assume that different quotes represent transfers that may have differing requirements regarding KYC and fiat accounts.
+To support this, the quote endpoints must return information about KYC and fiat account schemas that are required in order to actually perform a transaction for the requested quote.
 
 Quotes may vary depending on geo, and certain specific transfers may be unavailable in certain regions entirely. As such, we assume that the
 client will interpret a 200 response from this endpoint as verification that the provider is supported for the particular transfer parameters,
@@ -756,6 +762,15 @@ Some non-200s may be recovered from by modifying the transfer parameters; others
 The `POST /quote/in` endpoint is used to retrieve quotes used for transfers in to crypto from fiat currencies. In addition to returning quote information, it also
 returns the permissable types of KYC that a user must have on file to initiate the corresponding transfer, as well as the fiat account types that are allowed to be
 used for the transfer.
+
+Transfers in are notably different from transfers out, since they require the user to send fiat funds to a provider in exchange for crypto. Certain providers, or certain
+types of fiat accounts may have restrictions on *how* these fiat funds are transferred from the user to the provider. While some types of fiat accounts may support
+*debiting* the user's fiat account (i.e., initiating the transfer of fiat funds on behalf of the user), other types of fiat accounts may not support this behavior, and require
+the user themselves to initiate the transfer of fiat funds from their account to ther provider's.
+
+Certain types of fiat accounts may support *both* methods of fiat transfer, either initiated by the provider or the user. For such accounts, providers must choose one of these
+methods of transfer when returning the `/quote/in` response. By "default", it is assumed that a supported fiat account for a transfer in will be debited *by the provider*.
+Fiat account schema data in `/quote/in` responses may be augmented with additional information in order to denote that a transfer requires user action.
 
 ##### 3.4.1.1.1. Parameters
 
@@ -812,6 +827,7 @@ On success, the server MUST return an HTTP `200`, with the following response bo
 		[FiatAccountTypeEnum]: {
 			fiatAccountSchemas: {
 				fiatAccountSchema: `FiatAccountSchemaEnum`,
+				userActionType?: `TransferInUserActionTypeEnum`
 				allowedValues: {
 					[string]: `string[]`
 				}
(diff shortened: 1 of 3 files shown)


cajubelt:
i'm curious if there are some fiat account types where the provider doesn't actually need to know any information about the user's fiat account before a user-initiated fiat transfer takes place. For instance, for PSE do they actually need the account number for the originating account first?

fiatconnect-api.md
```diff
@@ -2211,6 +2264,30 @@ An enum listing the frequency, or how often, a particular fee needs to be paid.
 ]
 ```
 
+### 9.2.8. `TransferInUserActionTypeEnum`
```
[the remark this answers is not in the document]
jophish: do you suggest that the "action type" associated with each details schema just be a string literal equal to the schema name? that would mean, in order to specify that a transfer uses PSE, you'd need to return `userActionType: "PSEUserActionDetailsSchema"` which feels _slightly_ strange.

actually, we could do something like we do with KYC schemas currently; just give the action details schemas simpler names, e.g 'PSEUserAction'. in the types repo, we'd have a `UserActionSchema` (like with [KYC schemas here](https://github.com/fiatconnect/fiatconnect-types/blob/main/src/kyc.ts#L26)) which would be an enum like:

```typescript
enum UserActionSchema {
  PSEUserAction = 'PSEUserAction',
  IBANUserAction = 'IBANUserAction'
}

const userActionSchemaSchema = z.nativeEnum(UserActionSchema)
```

in addition, we'd also have a `UserActionSchemasSchema`, which would map from `UserActionSchema` enum values to the associated zod types (like we do [here](https://github.com/fiatconnect/fiatconnect-types/blob/main/src/kyc.ts#L121)), e.g.

```typescript
const userActionSchemasSchema = z.object(
  {
    [userActionSchema.enum.PSEUserAction]: pseUserActionZodSchema,
    [userActionSchema.enum.IBANUserAction]: ibanUserActionZodSchema,  
  }
)
```

okay, sorry for rambling -- you've convinced me :laughing:

fiatconnect-api.md
```diff
@@ -757,6 +763,15 @@ The `POST /quote/in` endpoint is used to retrieve quotes used for transfers in t
...
 used for the transfer.
 
+Transfers in are notably different from transfers out, since they require the user to send fiat funds to a provider in exchange for crypto. Certain providers, or certain
+types of fiat accounts may have restrictions on *how* these fiat funds are transferred from the user to the provider. While some types of fiat accounts may support
+*debiting* the user's fiat account (i.e., initiating the transfer of fiat funds on behalf of the user), other types of fiat accounts may not support this behavior, and require
+the user themselves to initiate the transfer of fiat funds from their account to ther provider's.
+
+Certain types of fiat accounts may support *both* methods of fiat transfer, either initiated by the provider or the user. For such accounts, providers must choose one of these
+methods of transfer when returning the `/quote/in` response. By "default", it is assumed that a supported fiat account for a transfer in will be debited *by the provider*.
+Fiat account schema data in `/quote/in` responses may be augmented with additional information in order to denote that a transfer requires user action.
```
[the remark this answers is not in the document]
jophish: yeah, this wasn't very clear; updated to be more readable.

fiatconnect-api.md
```diff
@@ -890,9 +906,13 @@ On success, the server MUST return a mapping from fiat account types to lists of
...
-This is identical in purpose and function to the `allowedValues` field for KYC schemas, discussed earlier.
+about the corresponding fiat account type. Each object MUST contain a `fiatAccountSchema` field representing the actual fiat account schema that can be used, an
+optional `allowedValues` field, and an optional `userActionType` field. The `allowedValues` object is an optional mapping from any number of keys in the selected
+fiat account schema to values that are allowed for that key. This is identical in purpose and function to the `allowedValues` field for KYC schemas, discussed earlier.
+`userActionType`, if present, denotes that this account schema, if used for the transfer, will require user action in order to transfer fiat funds from the user
+to the provider. Different values of `TransferInUserActionTypeEnum` denote different semantics with respect to what *kind* of action the user will have to perform
+in order to send fiat funds to the provider. If a fiat account schema requiring user action is used to execute a transfer with a compatible quote, the semantics
+of the `/transfer/in` endpoint will from normal, as discussed later.
```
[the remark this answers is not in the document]
jophish: ty -- removed "from normal" since i agree, it's not very helpful.

fiatconnect-api.md
```diff
@@ -757,6 +763,15 @@ The `POST /quote/in` endpoint is used to retrieve quotes used for transfers in t
 returns the permissable types of KYC that a user must have on file to initiate the corresponding transfer, as well as the fiat account types that are allowed to be
 used for the transfer.
 
+Transfers in are notably different from transfers out, since they require the user to send fiat funds to a provider in exchange for crypto. Certain providers, or certain
+types of fiat accounts may have restrictions on *how* these fiat funds are transferred from the user to the provider. While some types of fiat accounts may support
+*debiting* the user's fiat account (i.e., initiating the transfer of fiat funds on behalf of the user), other types of fiat accounts may not support this behavior, and require
+the user themselves to initiate the transfer of fiat funds from their account to ther provider's.
+
+Certain types of fiat accounts may support *both* methods of fiat transfer, either initiated by the provider or the user. For such accounts, providers must choose one of these
+methods of transfer when returning the `/quote/in` response. By "default", it is assumed that a supported fiat account for a transfer in will be debited *by the provider*.
```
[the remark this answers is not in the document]
jophish: Good point, and I like that phrasing much better -- updated to include it.

fiatconnect-api.md
```diff
@@ -1496,7 +1516,9 @@ the risk of creating a duplicate transaction. Idempotency keys MUST be implement
 
 #### 3.4.4.1. `POST /transfer/in`
 
-The `POST /transfer/in` endpoint is used to initiate a new transfer in from fiat to crypto.
+The `POST /transfer/in` endpoint is used to initiate a new transfer in from fiat to crypto. Since the transfer of fiat funds from a user to a provider
```
0xtuytuy, line 1519: This is more a comment than anything, but the /transfer/in could be irrelevant in a user-initiated transfer world. With some CICO partners, the transfer details will always remain constant for each user, meaning that the user could simply send money to the transfer details and the onramp should follow.

To not change the specs too much, i think we can keep this, but just pointing out that if the transfer details of Fiat are always the same, technically the partner does not need to call this endpoint to `initiate` the onramp.

fiatconnect-api.md
```diff
@@ -2439,7 +2504,50 @@ If `keyType` is `PHONE`, `key` MUST be an 11-digit [Brazilian mobile phone numbe
...
+#### 9.3.3.3. `PSEUserAction`
+
+`PSEUserAction` is a User Action Details Schema for transfers in requiring use of the Colombian [PSE payment system](https://www.pse.com.co/persona).
+
+```
+{
+	userActionType: `TransferInUserActionDetailsEnum.PSEUserAction`,
+	url: `string`,
+}
+```
```
0xtuytuy, line 2548: Should probably add one more here for FSP (Faster Payment) a UK payment scheme where IBAN is not used but rather Account Number and Sort code, the Schema could look like this:
```{
	userActionType: `TransferInUserActionDetailsEnum.IBANUserAction`,
	accountNumber: `string`,
	sortCode: `string`
}```

This is described here on the governments website: https://design-system.service.gov.uk/patterns/bank-details/

File: package.json [hunk 7]
```diff
@@ -1,6 +1,6 @@
 {
   "name": "specification",
-  "version": "1.0.3",
+  "version": "1.1.3",
```
0xtuytuy, line 3: 🚀

swagger.yaml
```diff
@@ -101,6 +101,17 @@ definitions:
       - IBANNumber
       - IFSCAccount
       - PIXAccount
+  UserActionTypeEnum:
+    type: "string"
+    enum: &UserActionTypeEnum
+      - PIX
+      - IBAN
+      - PSE
```
0xtuytuy, line 109: need to add FSP

0xtuytuy (commented):
Just a few comments

fiatconnect-api.md
jophish: Will you require FSP for user-initiated cash-ins? If you'll require this after this PR is merged, I'm fine with including this here, but if this is more speculatively useful, I'd rather not include it until someone definitely has a need for it.

swagger.yaml
jophish: see above comment

fiatconnect-api.md
```diff
@@ -1878,7 +1924,8 @@ though these are rough guidelines. Once a server has received an HTTP `200` stat
 ### 5.1.2. `WebhookTransferInStatusEventSchema`
 
 `WebhookTransferInStatusEventSchema` is the schema that defines webhook payloads for transfer in events. Note that this schema is *identical* to the
-one returned from the `GET /transfer/:transferId/status` endpoint.
+one returned from the `GET /transfer/:transferId/status` endpoint. The semantics for fields and their inclusion are also identical to those of the
+`GET /transfer:transferId/status` endpoint.
```
cajubelt, line 1928:
```suggestion
`GET /transfer/:transferId/status` endpoint.
```

fiatconnect-api.md
```diff
@@ -1891,15 +1938,18 @@ one returned from the `GET /transfer/:transferId/status` endpoint.
...
+	userActionDetails?: `TransferInUserActionDetailsEnum`
 }
 ```
 
 ### 5.1.3. `WebhookTransferOutStatusEventSchema`
 
 `WebhookTransferOutStatusEventSchema` is the schema that defines webhook payloads for transfer out events. Note that this schema is nearly identical to the
-one returned from the `GET /transfer/:transferId/status` endpoint; it lacks the optional `txHash` field, since it is not relevenat for transfers out.
+one returned from the `GET /transfer/:transferId/status` endpoint; it lacks the optional `txHash`, `userActionType` and `userActionDetails` field,
+since these are not relevant to transfers out. The semantics for fields and their inclusion are otherwise identical to those of the `GET /transfer:transferId/status`
```
cajubelt, line 1951:
```suggestion
since these are not relevant to transfers out. The semantics for fields and their inclusion are otherwise identical to those of the `GET /transfer/:transferId/status`
```

cajubelt (approved):

PR #98, opened 2023-06-25
fix(swagger): Update user action detail enum in swagger spec

jophish:
The user action details enum in the swagger spec differs from the actual specification, and thus the most recent changes to the fiatconnect-types repo. During testing, this was causing validations to fail where they shouldn't.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "specification",
-  "version": "1.1.3",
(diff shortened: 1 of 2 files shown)


cajubelt (approved):

cajubelt (commented):
might be worth incrementing the patch version

PR #102, opened 2023-11-05
feat(user-action-details): add AccountNumberUserAction

cajubelt:
We've seen FiatConnect providers use the URLUserActionDetails schema and direct users to their own website just to explain to them that they need to send funds to the same CICO provider's bank account. This defies the intention of the UserActionDetails feature, which is supposed to give FiatConnect clients a chance to explain to users what steps are needed to complete a transfer in-- not just hand them off to complete arbitrary steps on some other website. This PR adds a UserActionDetails schema that should bridge the gap, allowing CICO providers to solicit a bank transfer from end-users without redirecting them to another website.

### DIFF
--- a/fiatconnect-api.md
+++ b/fiatconnect-api.md
@@ -246,7 +246,8 @@
       - [9.3.3.1. `PIXUserAction`](#9331-pixuseraction)
       - [9.3.3.2. `IBANUserAction`](#9332-ibanuseraction)
       - [9.3.3.3. `PSEUserAction`](#9333-pseuseraction)
-      - [9.3.3.4. `URLUserAction`](#9333-urluseraction)
+      - [9.3.3.4. `URLUserAction`](#9334-urluseraction)
+      - [9.3.3.5. `BankAccountTransferUserAction`](#9335-bankaccounttransferuseraction)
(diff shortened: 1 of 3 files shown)


fiatconnect-api.md
```diff
@@ -2569,6 +2570,29 @@ schema.
...
+`AccountNumberUserAction` is a User Action Details Schema for transfers in which require the user to send funds to 
+an account identified by some number to complete the transfer of fiat funds. 
+
+```
+{
+	userActionType: `TransferInUserActionDetailsEnum.AccountNumberUserAction`,
+	institutionName: `string`,
+	accountName: `string`,
+	accountNumber: `string`,
+	transactionReference: `string`,
```
jophish, line 2584: Can this be made optional for providers that don't require this? I imagine many providers will be able to cross reference the account the payment is from with the one on file. You can add the provision that if it's present, it MUST be included in the actual transfer itself.

jophish (approved):
Optional comment, but A'ing since it's not blocking imo

fiatconnect-api.md
cajubelt: sounds reasonable. updated

